PR #56, opened 2020-08-13
Fix/decode resource names

VirginiaBalseiro:
This PR fixes:
- Resource names where being extracted from the URI without decoding, which meant they retained encoded characters ("Hello%20World").
- This PR fixes that by decoding the URI segment before returning the resource name for display (see screenshot below).

image


kevin-inrupt:
UX is correct. (Gotta love the Vercel Preview URL)

PR #60, opened 2020-08-20
Bumped solid-client-authn-browser to version 0.1.1

NSeydoux:
This upgrade is an attempt to fix the failure observed when logging in https://broker.demo-ess.inrupt.com.

This PR fixes bug #.


ajacksified:
I'm getting an error:

Screen Shot 2020-08-20 at 6 50 04 AM

NSeydoux:
Another bug still needs to be resolved in `solid-client-authn-browser`, where the first login fails and only succeeds once the cookie to the identity provider is set, but at least it should now be possible to log in, with a bad user experience.

PR #62, opened 2020-08-25
Feat/overwrite confirmation

VirginiaBalseiro:
# Display confirmation dialog when trying to overwrite a file
- Reused the `confirmationDialog` component
- Upon uploading a file, the `addFileButton` component now checks that the file already exists by trying to fetch its info (`fetchResourceInfoWithAcl` - TODO: change to `fetchResourceInfo`)

# Checklist

Note: this PR is a work in progress. Due to authentication errors I can throughly test it works and would like to get feedback and help with the authentication errors before attempting to merge this.

### DIFF
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -19,48 +19,174 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { useContext, useState } from "react";
+import { useContext, useState, useEffect } from "react";
 import T from "prop-types";
-import { saveFileInContainer } from "@inrupt/solid-client";
+import { overwriteFile, fetchResourceInfoWithAcl } from "@inrupt/solid-client";
 import { Button } from "@material-ui/core";
-import PodLocationContext from "../../src/contexts/podLocationContext";
 import SessionContext from "../../src/contexts/sessionContext";
+import PodLocationContext from "../../src/contexts/podLocationContext";
 import AlertContext from "../../src/contexts/alertContext";
+import ConfirmationDialogContext from "../../src/contexts/confirmationDialogContext";
 
-export default function AddFileButton({ onSave }) {
-  const { currentUri } = useContext(PodLocationContext);
-  const { session } = useContext(SessionContext);
-  const [isUploading, setIsUploading] = useState(false);
-  const { setMessage, setSeverity, setAlertOpen } = useContext(AlertContext);
+export function handleSaveResource({
+  fetch,
+  currentUri,
+  onSave,
+  setIsUploading,
+  setAlertOpen,
+  setMessage,
+  setSeverity,
+}) {
+  return async (uploadedFile) => {
+    try {
+      const response = await overwriteFile(
+        `${currentUri + uploadedFile.name}`,
+        uploadedFile,
+        { type: uploadedFile.type, fetch }
+      );
 
-  const onFileSelect = async (e) => {
-    if (e.target.files.length) {
-      setIsUploading(true);
+      onSave();
 
-      const [file] = e.target.files;
-
-      try {
-        const response = await saveFileInContainer(currentUri, file, {
-          fetch: session.fetch,
-          slug: file.name,
-        });
-
-        onSave();
-
-        setSeverity("success");
-        setMessage(
-          `Your file has been saved to ${response.internal_resourceInfo.sourceIri}`
-        );
-        setAlertOpen(true);
-      } catch (error) {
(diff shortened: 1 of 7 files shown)


components/addFileButton/index.jsx
```diff
@@ -19,48 +19,174 @@
...
+  onSave,
+  setIsUploading,
+  setAlertOpen,
+  setMessage,
+  setSeverity,
+}) {
+  return async (uploadedFile) => {
+    try {
+      const response = await overwriteFile(
+        `${currentUri + uploadedFile.name}`,
```
megoth, line 43: No need to use template strings for simple concatenations like these.

components/addFileButton/index.jsx
```diff
@@ -19,48 +19,174 @@
...
+      setAlertOpen(true);
+    } finally {
+      setIsUploading(false);
+    }
+  };
+}
+
+export async function findExistingFile(path, filename) {
+  try {
+    const existingFile = await fetchResourceInfoWithAcl(`${path}/${filename}`);
```
megoth, line 67: No need to create a variable here when you simply return it next line.

megoth (changes requested):
I was not able to test this without reverting the authentication client (fa26d053e8b8c9d6ff0c1cd6d06655dadb251bf6).

I did get a 401 for the HEAD request that checked for the existance of file - upon that response, PodBrowser continued to upload my file.

I would argue that the upload should fail if a 401 or other faulty HTTP statuses return, with an error message notifying the user about the problem.

components/addFileButton/index.jsx
```diff
@@ -19,48 +19,174 @@
...
+    return existingFile;
+  } catch (err) {
+    return null;
```
megoth, line 70: Per my last comments - we need to handle this case better.

ajacksified:
When I tested uploading a file with the same name as an existing file, it overwrote rather than prompting.

megoth:
> When I tested uploading a file with the same name as an existing file, it overwrote rather than prompting.

I experienced the same.

I suspect it has something to do with the use of a callback function that returns a falsy value, which is interpreted as there being no file present.

I suggest making the logic a bit clearer/easier to follow, by removing the callback function and rather make the returned value a bit more complex, i.e. `findExistingFile(path, filename, onFetchError): boolean` to `findExistingFile(path, filename, onFetchError): [boolean, error]`.

components/addFileButton/index.jsx
```diff
@@ -19,48 +19,195 @@
...
-import { saveFileInContainer } from "@inrupt/solid-client";
+import { overwriteFile, fetchResourceInfoWithAcl } from "@inrupt/solid-client";
 import { Button } from "@material-ui/core";
-import PodLocationContext from "../../src/contexts/podLocationContext";
 import SessionContext from "../../src/contexts/sessionContext";
+import PodLocationContext from "../../src/contexts/podLocationContext";
 import AlertContext from "../../src/contexts/alertContext";
+import ConfirmationDialogContext from "../../src/contexts/confirmationDialogContext";
+
+export function handleSaveResource({
```
megoth, line 31: If I understand correctly, these methods are exposed only so that you can test them, right? Might want to add

```
/**
 * @internal
 */
```

to indicate that they are not part of any public API.

megoth (changes requested):
I'm getting 401s when I try to upload files now, but I think it's because of the problems with the authentication library. I can make it work when I downgrade to the old authentication library at least.

Requesting changes for now, to ensure that there's nothing wrong with how we're using the new authentication library. If the problem must be fixed in the new auth library, I'll approve this PR.

components/addFileButton/index.jsx
ajacksified: I don't think we use typedoc anywhere - throughout the codebase we're exporting little functions like these. I don't think it's particularly necessary to call out as internal.

megoth: True, and it's not like we're exposing PodBrowser as a library, so no need to indicate internal and public API. So ignore my previous suggestion ^_^

components/addFileButton/index.jsx
```diff
@@ -19,48 +19,195 @@
...
+      setAlertOpen(true);
+    } finally {
+      setIsUploading(false);
+    }
+  };
+}
+
+export async function findExistingFile(path, filename) {
+  try {
+    return await fetchResourceInfoWithAcl(path + filename);
```
ajacksified, line 72: You need to pass `fetch` here - should fix the 401s.

megoth (approved):

PR #63, opened 2020-08-27
Fixing drawer using Prism React Components

megoth:
# New feature description

Making use of the new repo Prism React Components to fix the drawer so that it looks closer to Andrei's designs. The new repo should make it easier to pull patterns into PodBrowser in general.

This does break the disabled style of the upload button, but this needs to be properly designed in any case.

# Checklist

- [x] All acceptance criteria are met.
- [ ] Relevant documentation, if any, has been written/updated.
- [ ] The changelog has been updated, if applicable.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #61, opened 2020-08-21
Implement functions to create a new AddressBook

daytonn:
# Create contacts list

This PR adds functions to handle creating new address book containers.

## Checklist

- [x] All acceptance criteria are met.
- [ ] Relevant documentation, if any, has been written/updated.
- [ ] The changelog has been updated, if applicable.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #64, opened 2020-08-27
Solidos 320/create contact

daytonn:
# Add a new contact by webId

This PR adds a `/contacts` and `/contacts/add` page that allows you to create a new contact by entering an agent's webId.

This includes a pretty heavy refactorying of `solidClientHelpers`. I split out the functionality into specific modules to clear up responsibilities and make things easier to understand.

Having grappled with the SDK and getting some more understanding around datasets, I've written some basic utilities to work with creating datasets locally that map objects to the sdk functions that can create datasets from those object. For example creating a contact looks something like this:

```js
const contact = {
  webId,
  addresses: [
    {
      countryName: "Fake Country",
      locality: "Fake Town",
      postalCode: "55555",
      region: "Fake State",
      streetAddress: "123 Fake St.",
    },
  ],
  fn: "Test Person",
  emails: [
    {
      type: "Home",
      value: "test@example.com",
    },
    {
      type: "Work",
      value: "test.person@example.com",
    },
  ],
  telephones: [
    {
      type: "Home",
      value: "555-555-5555",
    },
  ],
  organizationName: "Test Company",
  role: "Developer",
};
const addressBookIri = "https://user.example.com/contacts"
const { response, error } = await createContact(addressBookIri, contact);
```
Peek 2020-08-26 20-10

_Note: Sorry this is a pretty big one since undoing TS for `solidClientHelpers` meant changing a lot of consuming code_
# Checklist


ajacksified:
I assume this makes #61 obsolete?

daytonn:
Yeah, because I moved everything and did a heavy refactor, #61 is obsolete.

src/addressBook/index.js
```diff
@@ -0,0 +1,289 @@
+/**
```
daytonn, line 1: I think we may want to figure out if these are "model" files and where we want to formally put them. I'm not convinced `src/` is the right place for this file.

src/hooks/solidClient/index.test.js
```diff
@@ -147,35 +146,16 @@ describe("fetchResourceDetails", () => {
...
-    jest
-      .spyOn(solidClientHelpers, "fetchFileWithAcl")
-      .mockResolvedValue({ iri, types: ["type"], file: new Blob(["file"]) });
-
-    const resourceDetails = await fetchResourceDetails(iri);
-
-    expect(resourceDetails.name).toEqual("/file.txt");
-    expect(resourceDetails.iri).toEqual(iri);
-  });
 });
```
daytonn, line 149: This test was moved to `resource.test.js`

components/header/userMenu/index.tsx
```diff
@@ -37,55 +38,75 @@ export default function UserMenu(): ReactElement {
...
-          className={clsx(
-            bem("icon-user"),
-            bem("header-banner__aside-menu-trigger-icon")
-          )}
-          aria-label="User Menu"
-        />
-      </button>
+    <>
+      <div className={bem("header-banner__main-nav")}>
+        <Link href="/contacts" replace>
```
megoth, line 43: I think this needs to be replaced with the Main Menu pattern in Prism - but we can do it in another PR, as we probably want to use the Main Menu as a React component instead.

src/addressBook/index.js
megoth: Where should it be if not under `src/`?

megoth (changes requested):
The Contacts Pane in Data Browser also adds an entry for each contact added, e.g. `contacts/People/[unique id]/index.ttl`.

It should also have the `vcard:fn` term, which means that we have redundant information, something we need to discuss with Tim and the rest of the team.

megoth (changes requested):
Seems also that `this` is no longer added for contacts/index.ttl, which is needed for Data Browser to do its thing.

src/addressBook/index.js
daytonn: It will till be under `src` but I think we may want some other kind of sub-folder to contain these types of files. I just called out that I'm not sure that's where we ultimately want to keep these types of modules.

megoth: We need a discussion on how to conceptualize these modules, to see if there's a pattern that we can formalize into something consistent.

megoth: So yeah, I agree with you ^_^

components/container/styles.js
```diff
@@ -0,0 +1,48 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { createStyles } from "@solid/lit-prism-patterns";
+
+const styles = (theme) => {
+  return createStyles(theme, ["container", "table"], {
+    "container-view": {
```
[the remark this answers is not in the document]
megoth: You might want to change this now that https://github.com/inrupt/pod-browser/pull/63 is merged. Or is it better to take it in another PR

daytonn:
@megoth https://github.com/inrupt/pod-browser/pull/64/files#diff-20afc3b0749a86b6bcf17ac77bd5629cR57

megoth:
> @megoth https://github.com/inrupt/pod-browser/pull/64/files#diff-20afc3b0749a86b6bcf17ac77bd5629cR57


It doesn't seem to work properly, see the content of index.ttl that was created for me:

```
<#159895555679908723988270597993> a <http://www.w3.org/2006/vcard/ns#AddressBook>;
    <http://www.w3.org/ns/auth/acl#owner> <https://megoth-demo1.inrupt.net/profile/card#me>;
    <http://purl.org/dc/elements/1.1/title> "Contacts";
    <http://www.w3.org/2006/vcard/ns#nameEmailIndex> <https://megoth-demo1.inrupt.net/contacts/people.ttl>;
    <http://www.w3.org/2006/vcard/ns#groupIndex> <https://megoth-demo1.inrupt.net/contacts/groups.ttl>.
```

The subject here should be <#this>.

components/container/styles.js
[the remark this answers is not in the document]
megoth: I assume we'll do this in another PR (this PR is complex enough already).

megoth:
I'm getting some errors with the new auth client:

bilde

megoth:
(sorry, misclick)

megoth:
I tried out the latest updates, and still getting 401s, so reckon this is still work in progress?

I tested this out on ESS, and see that you assume that the domain is the Pod's root. That is not necessarily the case (e.g. https://ldp.demo-ess.inrupt.com/113069137442933403687/ is the root of my Pod), which means that using https://ldp.demo-ess.inrupt.com/contacts will fail with a 403.

megoth (changes requested):
I'm still getting a 401 when PB tries to create `<pod-root>/contacts`. Something I'm missing?

ajacksified (approved):

PR #69, opened 2020-09-02
Feat/create container

VirginiaBalseiro:
# New feature description

- Added a new Create Folder button with a flyout that allows the user to create a new empty folder
  - If folder name already exists, it appends a number: Folder(1) and so on.
- To avoid making many requests to see if the name exists, I am checking in the data passed from the parent component. In hindsight this approach could be used in the AddFileButton as well - good to keep in mind when we review that button's overwrite behavior as well.

# Checklist


components/container/index.tsx
```diff
@@ -115,7 +115,7 @@ export default function Container({ iri }: IPodList): ReactElement {
   /* eslint react/jsx-props-no-spreading: 0 */
   return (
     <>
-      <PageHeader mutate={mutate} />
+      <PageHeader mutate={mutate} data={data} />
```
megoth, line 118: A bit skeptic of passing variables way through multiple components like this, makes reusability a bit harder/cumbersome. Could perhaps the use of context be more approapriate? And perhaps call it something less ambigous than data?

megoth (changes requested):
I think the flyout should close after the container is created. (But please check this with Andrei.)

Also think reusability would be better with the use of context and passing the data that way, but open for it to be passed directly as well.

components/container/index.tsx
megoth: @daytonn do you have an opinion on this?

VirginiaBalseiro: @megoth I renamed the variable to `resourceList` - open to suggestions for a better name!
I also fixed the issue of the flyout not closing when saving, it does now.
Re: context - I decided against it since it is only one prop and only passing through 2 components. However if we are moving adding files to flyout instead of the current dialog, it might be the case that we can have a reusable component for both and a flyout context, but we can always revisit this when we do that work.

megoth (approved):

PR #72, opened 2020-09-07
Added method that gets or creates PodBrowser preferences

megoth:
# Getting PodBrowser preferences (or creates if missing)

Work in progress (draft)

# Checklist


megoth:
Currently some existing tests written in TS-files fails, probably because I've made the validation of TS files in tests more strict by adding transform explicitly in jest.config.js.

megoth:
This PR changes a lot of files, but most of them are very small changes, so I hope this is acceptable.

I've used mocked TTL files instead of mocking the result of Solid client and React methods (I would prefer we avoid mocking useState and useContext - this makes our tests very brittle). I think this is a better approach, but open for discussion as always.

components/addFolderFlyout/index.test.jsx
```diff
@@ -19,6 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import React from "react";
```
ajacksified, line 22: These should be auto-imported; was the linter complaining?

src/hooks/usePodBrowserSettings/index.js
```diff
@@ -0,0 +1,43 @@
...
+import { getOrCreateSettings } from "../../solidClientHelpers/settings";
+import SessionProvider from "../../contexts/sessionContext";
+
+export default function usePodBrowserSettings() {
+  const { session } = useContext(SessionProvider);
+  const profileInfo = useAuthenticatedProfile();
+  const [settings, setSettings] = useState(null);
+
+  useEffect(() => {
+    if (!profileInfo || !session) return;
```
ajacksified, line 34: `session` is always defined - you want `session.info.isLoggedIn` here.

components/addFolderFlyout/index.test.jsx
megoth: It seems to be Jest itself, giving me `ReferenceError: React is not defined`.

src/hooks/usePodBrowserSettings/index.js
```diff
@@ -0,0 +1,43 @@
...
+    getOrCreateSettings(profileInfo.webId, session)
+      .then((settingsUrl) =>
+        getSolidDataset(settingsUrl, { fetch: session.fetch })
+      )
+      .then(setSettings);
```
ajacksified, line 39: Imo, this would read more easily as an async:

```
useEffect(() => {
  (async function getSettings() {
    const settingsUrl = await getOrCreateSettings(profileInfo.webId, session);
    const settings = await getSolidDataset(settingsUrl, { fetch: session.fetch });
    setSettings(settings);
  })();
});
```

src/solidClientHelpers/settings.js
```diff
@@ -0,0 +1,130 @@
...
+    userSettings,
+    userSettingsResource,
+    userSettingsUrl,
+    session
+  );
+  return getOrCreatePodBrowserSettingsResource(podBrowserSettingsUrl, session);
+}
+
+// eslint-disable-next-line import/prefer-default-export
+export async function getOrCreateSettings(webId, session) {
```
ajacksified, line 127: Seems a little weird to me that `getOrCreateSettings` returns an iri instead of the settings dataset - I'd either rename this to `getOrCreateSettingsIri` or just return the dataset here.

ajacksified (changes requested):

src/hooks/usePodBrowserSettings/index.js
megoth: I agree, just didn't think we wanted to have async/await in the useEffect block.

Will change it.

ajacksified: It's situational - I tend not to if there's a single promise, but if there are multiple, I think it's easier to read.

megoth: After some changes there's only one promise, so I didn't use the async/await then.

ajacksified (approved):

ajacksified (approved):

PR #73, opened 2020-09-08
Add unique ids

VirginiaBalseiro:
# Adding ids to all interactive elements

This PRs adds ids to:

- Login components: inputs, buttons
- User Menu button and logout button
- Resource details: delete button, sharing and app permissions button, download button
- Breadcrumbs (list, not individual crumbs)
- Sort toggle buttons (File/Type)
- Header banner link
- Footer TOS link
- Folder Details and Upload File buttons

**Notes**:
* Elements that have to be modified on Prism:
  - Drawer close button
  - Page header title

Some elements used an `id` for other purposes, I did not add a `data-testid` for those.

# Checklist

### DIFF
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -27,6 +27,9 @@ import PodLocationContext from "../../src/contexts/podLocationContext";
 import AlertContext from "../../src/contexts/alertContext";
 import ConfirmationDialogContext from "../../src/contexts/confirmationDialogContext";
 
+const TESTCAFE_ID_UPLOAD_BUTTON = "upload-file-button";
+const TESTCAFE_ID_UPLOAD_INPUT = "upload-file-input";
+
 export function handleSaveResource({
   fetch,
   currentUri,
@@ -199,10 +202,14 @@ export default function AddFileButton({ className, onSave, resourceList }) {
 
   return (
     // eslint-disable-next-line jsx-a11y/label-has-associated-control
-    <label id="upload-file-button" disabled={isUploading} className={className}>
+    <label
+      data-testid={TESTCAFE_ID_UPLOAD_BUTTON}
+      disabled={isUploading}
+      className={className}
+    >
       {isUploading ? "Uploading..." : "Upload File"}
       <input
-        id="upload-file-input"
+        data-testid={TESTCAFE_ID_UPLOAD_INPUT}
         type="file"
         style={{ display: "none" }}
         onClick={(e) => {
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -199,9 +199,10 @@ export default function AddFileButton({ className, onSave, resourceList }) {
 
   return (
     // eslint-disable-next-line jsx-a11y/label-has-associated-control
-    <label disabled={isUploading} className={className}>
+    <label id="upload-file-button" disabled={isUploading} className={className}>
       {isUploading ? "Uploading..." : "Upload File"}
       <input
+        id="upload-file-input"
         type="file"
         style={{ display: "none" }}
         onClick={(e) => {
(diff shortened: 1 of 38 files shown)


VirginiaBalseiro:
Bringing this comment by @Vinnl on a related PR:

>  ideally, matching IDs in tests is the last resort, and even then, is on data-testid rather than on id, for the reasons mentioned in this article: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

@kevin-inrupt would this still work for you with TestCafe if we changed all the `id` to `data-testid`?

Opinions @megoth ?

megoth:
> Opinions @megoth ?

Always 😸 

I agree that `data-testid` are better than `id`, as it becomes more specific why it's there, and less chance of being confused with other uses of `id`. Also an improvement is that since the semantics are not limiting it to a unique element, we could use this instead of splitting up selectors between classes and IDs. Writing `[testid=SomeTestId]` instead of `#SomeTestId` is acceptable IMO.

So 👍 from me.

kevin-inrupt:
I was basing the use of IDs off of the recommendations in https://devexpress.github.io/testcafe/documentation/reference/test-api/selector/constructor.html#initialize-selectors.

Out-of-the-box, TestCafe does not support data-testid. There is a library (https://testing-library.com/docs/testcafe-testing-library/intro#usage) that we could use. Hence we could switch. 

I read through the blog post https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change (which incidentally was written by one of the contributors to the library). I agree selecting by classname is not the way to go, but not sure the difference between id and data-testid.

megoth:
> Out-of-the-box, TestCafe does not support data-testid.

But you can use any CSS selector in TestCafe, right? If so, you can use the attribute selectors instead of IDs, e.g. `[data-testid=SomeTestId]` instead of `#SomeTestId`.

ajacksified:
I don't like arbitrary strings, and I also don't like adding code just to make tests pass. Is it possible to use element selectors to get what you want? It'd be more resilient to code refactoring to use something like, "get the first button on the page (`button:first-of-type`)" rather than "get this arbitrary ID that someone will inevitably change or remove."

We also won't be able to use `id` attributes on any element meant to be used more than once, which can lead to difficult-to-debug issues.

If we _have_ to have `id`s, they should be statically exported from the top of the file as documentation:

`const ID_FOR_TESTCAFE_HOOKS_WHICH_MAKES_ME_SAD = "some-arbitrary-string"`

components/addFileButton/index.jsx
```diff
@@ -27,6 +27,9 @@ import PodLocationContext from "../../src/contexts/podLocationContext";
 import AlertContext from "../../src/contexts/alertContext";
 import ConfirmationDialogContext from "../../src/contexts/confirmationDialogContext";
 
+const TESTCAFE_ID_UPLOAD_BUTTON = "upload-file-button";
```
megoth, line 30: I think we should the constants into one file, that way we get one place to look up used IDs.

components/containerDetailsButton/index.tsx
```diff
@@ -33,6 +33,7 @@ export default function ContainerDetailsButton({
 }: Props): ReactElement {
   return (
     <ResourceLink
+      id="folder-details-button"
```
megoth, line 36: Think you forgot this one?

megoth (changes requested):
The problem with many of these IDs is that they won't be unique once we reuse the components. So we need a system where IDs are unique, even when they appear in reused components.

components/addFileButton/index.jsx
ajacksified: I don't agree - I think it'll be too much back and forth when developing. @kevin-inrupt's tool inspects the dom anyways, rather than browsing source code.

megoth (approved):

PR #78, opened 2020-09-11
Move mouse enter event to button to prevent infinite toggling

VirginiaBalseiro:
When clicking the user menu button and then hovering over "Log out", the toggle function was triggered causing back and forth toggling, which resulted in the "Log out" menu item being displayed intermittently.

Moving the mouse enter event to the button instead of the "Log out" menu item fixes this.

### DIFF
--- a/components/header/__snapshots__/index.test.tsx.snap
+++ b/components/header/__snapshots__/index.test.tsx.snap
@@ -867,7 +867,6 @@ font-display: block;",
           <UserMenu>
             <div
               className="PodBrowser-header-banner__aside-menu PodBrowser-header-banner__aside-menu--popup"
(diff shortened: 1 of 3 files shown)


megoth (changes requested):
This can quickly lead to the situation where the user menu stays open because state gets scrambled.

Screen Recording 2020-09-11 at 15 57 19

I suggest using throttle to give the user some time to navigate to the user menu before it closes (maybe 300ms).

VirginiaBalseiro:
> I suggest using throttle to give the user some time to navigate to the user menu before it closes (maybe 300ms).

if we add throttle/debounce and don't move the mouseEnter to the button, it still flickers, it just flickers more slowly (edited)

PR #80, opened 2020-09-14
Feature/no control access error

megoth:
# New feature description: Created page Access Required and hook useRedirectIfNoControlAccessToPod

Hook `useRedirectIfNoControlAccessToPod` is added to pages that requires access to your pod, i.e. Home and Resource.

This builds upon the work in https://github.com/inrupt/pod-browser/pull/72 in order to mock TTL files.

# Checklist


components/pages/access-required/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/* eslint react/jsx-one-expression-per-line: 0 */
```
ajacksified, line 22: Why? If this is causing issues, we should revisit the rule.

components/pages/access-required/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+import styles from "./styles";
+import { useRedirectIfLoggedOut } from "../../../src/effects/auth";
+
+const useStyles = makeStyles((theme) => createStyles(styles(theme)));
+
+export default function AccessRequired() {
+  useRedirectIfLoggedOut();
+
+  const bem = useBem(useStyles());
+  const [appUrl, setAppUrl] = useState("http://podbrowser.inrupt.com/");
```
ajacksified, line 37: Use an empty string here rather than showing a potentially wrong value.

components/pages/access-required/styles.js
```diff
@@ -0,0 +1,34 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { createStyles } from "@solid/lit-prism-patterns";
+
+export default function styles(theme) {
+  return createStyles(theme, [], {
+    title: {
```
ajacksified, line 26: This style seems like something that belongs in prism, rather than as a one-off.

pages/access-required.jsx
```diff
@@ -0,0 +1,27 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import React from "react";
+import AccessRequired from "../components/pages/access-required";
+
+export default function Custom404() {
```
ajacksified, line 25: `Custom404` -> `AccessRequired`

src/effects/auth/index.js
```diff
@@ -80,3 +80,28 @@ export function useRedirectIfLoggedIn(location = "/"): void {
...
+  location = "/access-required"
+) {
+  const { session, isLoadingSession } = useContext(SessionContext);
+  const profile = useAuthenticatedProfile();
+
+  useEffect(() => {
+    if (isLoadingSession || !profile) return;
+
+    Promise.all(
+      profile.pods.map((pod) =>
```
ajacksified, line 94: Rather than reading all pods for a profile, we should look at the current pod (especially in the case that we're visiting someone else's pod.) Use `PodLocationContext` for the current iri, and get its dataset.

ajacksified (changes requested):

components/pages/access-required/index.jsx
megoth: Prettier and ESLint have a conflict with <strong>{appUrl}</strong>. I've added a comment about this in the latest change.

src/effects/auth/index.js
megoth: But we should be able to visit other Pods without having Control access, no?

I agree that we should have some access check for pods you're visiting, but I thought this PR was suppose to check that you have Control access to the Pods you have listed in your profile?

ajacksified: Even if that's true, we shouldn't be reading the pod iri from the profile. This should be checking the current pod.

That said, we should only run `useRedirectIfNoControlAccessToPod` if you're on your own pod, which would require checking the current pod iri against the list from your profile.

components/pages/access-required/styles.js
megoth: Yes, would be better for reusability. Not quite sure how to name it though, any suggestions? Something along the lines of info-title? (To be reused for titles on more static pages, to convey some info to the users.)

megoth: We already have some rules for titles in general, so didn't want to adjust [the default typography that was created based on the earlier review by Andrei](https://www.figma.com/file/W2o8rM7yu5WxuRAlAOukuG/Pod-Browser?node-id=204%3A503).

src/effects/auth/index.js
megoth: Ok, so I'll change the logic to the following:

* `useRedirectIfNoControlAccessToOwnPod`: This will kick in _if_ you're visiting your own Pod (i.e. one of the ones listed in your profile) _and_ you don't have Control access
* `useRedirectIfNoAccessToPod`: This will kick in _if_ you're trying to visit any Pod that you don't have Read access to

Does that sound reasonable?

ajacksified: Yes, that sounds reasonable.

components/pages/access-required/styles.js
ajacksified: I've seen `hero` used for this kind of top banner (like a bigger h1) in bootstrap.

megoth: Introduced Hero element in Prism. If https://github.com/inrupt/prism/pull/95 is accepted, I'll make sure to build a new version, and update this PR.

src/effects/auth/index.js
megoth: Think it does what you expect now.

ajacksified (approved):

PR #88, opened 2020-09-16
Feat/contacts sort

VirginiaBalseiro:
# Sort contacts
In this PR:
 - Refactor Contacts List to use `solid-ui-react` sortable Table component 
 - Updated and added new tests

# Screenshot
image

# Checklist


components/contactsList/index.jsx
```diff
@@ -134,24 +137,44 @@ function ContactsList() {
...
-              </tr>
-            ))}
-          </tbody>
-        </table>
+        <Table
+          things={contacts}
+          className={bem("table")}
+          // prettier-ignore
+          ascIndicator={(
+            <span role="img" aria-label="Sorted in ascending order">
```
ajacksified, line 145: I think it'd be nice to pull the indicators into their own components, so we can re-use them as we build more tables.

components/contactsList/index.jsx
```diff
@@ -134,24 +137,44 @@ function ContactsList() {
...
+            </span>
+          )}
+        >
+          <TableColumn
+            property={hasPhotoPredicate}
+            header=""
+            datatype="url"
+            body={({ value }) => {
+              return (
+                <Avatar
```
ajacksified, line 164: Note for later: we'll eventually want to replace `Avatar` with our own component that applies prism avatar classes to the react SDK `<Image>` component, which would allow us to make authenticated requests.

ajacksified (changes requested):

components/contactsList/styles.js
```diff
@@ -19,10 +19,23 @@
...
-import { createStyles } from "@solid/lit-prism-patterns";
+import { createStyles, table } from "@solid/lit-prism-patterns";
 
 const styles = (theme) => {
-  return createStyles(theme, ["container", "table"], {
+  const tableStyles = table.styles(theme);
+  return createStyles(theme, ["container", "table", "icons"], {
+    table: {
+      ...tableStyles.table,
+      "& tbody": tableStyles.table__body,
```
megoth, line 29: I realize now that we might just as well make this default behavior in Prism (e.g. have rules for both `table__body` and `& tbody` that points to the same rules). That way we don't have to repeat this block of code for every table in PB.

(I initially don't want to do this, as it makes CSS specificity harder for later overrides, but I think it's preferrable in this case.)

src/addressBook/index.test.js
```diff
@@ -247,27 +302,75 @@ describe("getPeople", () => {
       ]);
 
     jest
-      .spyOn(solidClientFns, "getStringNoLocale")
-      .mockReturnValueOnce(expectedPerson1.nickname)
-      .mockReturnValueOnce(expectedPerson1.name)
-      .mockReturnValueOnce(expectedPerson2.nickname)
-      .mockReturnValueOnce(expectedPerson2.name);
+      .spyOn(solidClientFns, "addStringNoLocale")
```
megoth, line 305: Could we try to pass turtle files and let Solid client do its thing, instead of mocking the response from the various Solid Client methods?

src/addressBook/index.test.js
```diff
@@ -247,27 +302,75 @@ describe("getPeople", () => {
...
+      })
+      .mockResolvedValueOnce({
+        response: { dataset: "Profile 1", iri: expectedPerson1.webId },
+      })
+      .mockResolvedValueOnce({
+        response: { dataset: "Profile 2", iri: expectedPerson2.webId },
+      });
+
+    jest
+      .spyOn(solidClientFns, "getUrlAll")
```
megoth, line 354: Again, I would prefer that we avoid this mocking pattern.

src/addressBook/index.test.js
megoth: Alternatively, use the mocking methods that Solid client brings.

components/contactsList/index.jsx
megoth: I second this - it should be a small component in Prism React components.

megoth (changes requested):

components/contactsList/index.jsx
megoth: Should we create a ticket for this as a reminder to ourselves?

src/addressBook/index.test.js
megoth: Because of all the bugs we've found elsewhere, we'll defer changes to mocking.

src/addressBook/index.js
```diff
@@ -141,21 +141,14 @@ export async function getPeople(containerIri, fetch) {
   const profiles = profileResponses
     .filter(({ error: e }) => !e)
     .map(({ response }) => response)
+    .filter(({ iri }) => iri.includes("/profile/card#me")) // filtering by valid WebId so it doesn't error out if the user entered a contact with an invalid ID
```
megoth, line 144: WebIds are not necessarily following the form `/profile/card#me`, that's just a form that NSS uses.

(If anything, we need to resolve "bad" webIds when we fetch them. But we should create another ticket to handle those situations.)

megoth (changes requested):

megoth (approved):

ajacksified (approved):

PR #94, opened 2020-09-23
Feat/create bookmark

VirginiaBalseiro:
# Create bookmark

- Added a new Bookmark component that uses bookmarks => `useBookmark` checks if Bookmarks already exists or creates it at `/bookmarks/index.ttl`
- Bookmarks is kept in a context to avoid drilling it down several layers since the Bookmark icon is inside container > row component since it is used in every operation (add, remove, display selected/unselected,etc).
- when clicking on the star icon, it checks if the bookmark already exists and if so, it does nothing [we will add delete functionality later], if not, it adds the bookmark.

# Checklist


src/solidClientHelpers/bookmarks.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export async function addBookmark(bookmarkIri, bookmarks, fetch) {
+  const { dataset, iri } = bookmarks;
+  const bookmarkTitle = getResourceName(bookmarkIri);
+  const bookmark = defineThing(
+    {},
+    (t) => addUrl(t, rdf.type, "http://www.w3.org/2002/01/bookmark#Bookmark"),
+    (t) =>
+      addStringNoLocale(t, "http://purl.org/dc/terms/title", bookmarkTitle),
```
megoth, line 52: There's also the terms vocabulary in the "rdf-namespaces" package, let's use it.

megoth (changes requested):
One minor change, and it's good to go ^_^

components/bookmark/index.jsx
```diff
@@ -0,0 +1,124 @@
...
+      setBookmarked(true); // optimistically set bookmarked to true so it doesn't take that long to display the item as bookmarked
+      setBookmarks(response);
+    }
+  };
+};
+
+const isBookmarked = (iri, dataset) => {
+  const listOfRecallsUrls = getUrlAll(
+    dataset,
+    "http://www.w3.org/2002/01/bookmark#recalls"
```
ajacksified, line 66: This (and all property IRIs) should be an exported constant.

components/bookmark/index.jsx
```diff
@@ -0,0 +1,124 @@
...
+  );
+  if (!listOfRecallsUrls) {
```
ajacksified, line 68: The solid-client-js api looks like it [always returns an array](https://docs.inrupt.com/developer-tools/api/javascript/solid-client/modules/_thing_get_.html#geturlall); you should be able to skip this check and just use the `includes` below.

megoth (approved):

PR #96, opened 2020-09-24
Feat/remove bookmark

VirginiaBalseiro:
# Add functionality to remove bookmark

Clicking on a previously bookmarked resource's bookmark icon will now remove the bookmark.

## Notes
To prevent errors from a user clicking too fast, the button is temporarily disabled until the bookmarks' dataset is effectively updated. 

# Checklist

### DIFF
--- a/components/bookmark/index.jsx
+++ b/components/bookmark/index.jsx
@@ -30,6 +30,7 @@ import styles from "./styles";
 import {
   addBookmark,
   RECALLS_PROPERTY_IRI,
+  removeBookmark,
 } from "../../src/solidClientHelpers/bookmarks";
 import BookmarksContext from "../../src/contexts/bookmarksContext";
 import AlertContext from "../../src/contexts/alertContext";
@@ -38,26 +39,31 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
 export const toggleBookmarkHandler = ({
   bookmarks,
-  bookmarked,
   iri,
   setSeverity,
   setMessage,
   setAlertOpen,
-  setBookmarked,
+  bookmarked,
   setBookmarks,
+  setBookmarked,
+  setDisabled,
   fetch,
 }) => {
   return async () => {
+    setDisabled(true);
+    let results;
     if (bookmarked) {
-      return; // returning for now until we have the remove logic
+      results = await removeBookmark(iri, bookmarks, fetch);
+    } else {
+      results = await addBookmark(iri, bookmarks, fetch);
     }
-    const { response, error } = await addBookmark(iri, bookmarks, fetch);
+    const { response, error } = results;
     if (error) {
       setSeverity("error");
       setMessage(error);
       setAlertOpen(true);
     } else {
-      setBookmarked(true); // optimistically set bookmarked to true so it doesn't take that long to display the item as bookmarked
+      setBookmarked(!bookmarked);
       setBookmarks(response);
     }
   };
@@ -75,6 +81,7 @@ export default function Bookmark({ iri }) {
   const bem = useBem(useStyles());
   const { bookmarks, setBookmarks } = useContext(BookmarksContext);
   const [bookmarked, setBookmarked] = useState();
+  const [disabled, setDisabled] = useState(false);
 
   useEffect(() => {
     if (!bookmarks) {
@@ -83,6 +90,7 @@ export default function Bookmark({ iri }) {
     const { dataset } = bookmarks;
     const bookmarkState = isBookmarked(iri, dataset);
     setBookmarked(bookmarkState);
+    setDisabled(false);
   }, [bookmarks, iri]);
 
   const toggleBookmark = toggleBookmarkHandler({
@@ -94,6 +102,7 @@ export default function Bookmark({ iri }) {
     setAlertOpen,
     setBookmarked,
     setBookmarks,
+    setDisabled,
     fetch,
   });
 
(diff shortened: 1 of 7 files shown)


components/bookmark/index.jsx
```diff
@@ -38,26 +39,31 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
...
     }
-    const { response, error } = await addBookmark(iri, bookmarks, fetch);
+    const { response, error } = results;
     if (error) {
       setSeverity("error");
       setMessage(error);
       setAlertOpen(true);
     } else {
-      setBookmarked(true); // optimistically set bookmarked to true so it doesn't take that long to display the item as bookmarked
+      setBookmarked(!bookmarked);
```
ajacksified, line 66: To set this optimistically, should you do it above the `await` instead, and revert the change on failure?

VirginiaBalseiro: It is "half-optimistic" right now actually, because it doesn't wait until the bookmarks dataset is updated, just that it doesn't throw an error vs. not setting it here and having to wait both the response from the `await` and the update of the bookmarks. If I don't wait until the response is back, there is some flickering while it's waiting for the bookmarks to be updated.

megoth: I think updating the state optimistally + not allowing further changes until updated state is resolved (i.e. disable button) might be the way to go on this?

components/bookmark/index.jsx
```diff
@@ -38,26 +39,31 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
 export const toggleBookmarkHandler = ({
```
megoth, line 40: In general I would prefer if we stop exporting functionality that should be internal like this. I understand that it's easier to test functionality this way, but we should be able to test the input and output of the component without having to destructure the inner working like this.

This also creates a hellish parameter to maintain, and a lot of unnecessary wiring.

Component files should only export one function, the component itself.

src/solidClientHelpers/bookmarks.js
```diff
@@ -49,19 +52,43 @@ export const BOOKMARK_TYPE_IRI = "http://www.w3.org/2002/01/bookmark#Bookmark";
...
-    (t) => addUrl(t, rdf.type, BOOKMARK_TYPE_IRI),
-    (t) => addStringNoLocale(t, dct.title, bookmarkTitle),
-    (t) => addUrl(t, RECALLS_PROPERTY_IRI, bookmarkIri),
-    (t) => addDatetime(t, dct.created, new Date())
+  const [existingBookmark] = getThingAll(dataset).filter(
+    (t) => getUrl(t, RECALLS_PROPERTY_IRI) === bookmarkIri
   );
+  let results;
+  if (existingBookmark) {
+    results = { response: bookmarks, error: null };
```
megoth, line 60: Should we somehow notify the user that the bookmark already existed? Or is this a very edge-case?

megoth (changes requested):
I would like you to make `toggleBookmarkHandler` internal in ` components/bookmark/index.jsx`, and change the tests to reflect this.

If you're unable to do that, or think it should be moved into a bigger discussion with the team, I'm ok with the current changes and can approve the PR.

components/bookmark/index.jsx
VirginiaBalseiro: The button is disabled until the bookmarks are updated. The problem is that there is a brief moment where the bookmarks are updated but it hasn't checked if the bookmark already exists or not.

src/solidClientHelpers/bookmarks.js
VirginiaBalseiro: The problem is that if the bookmark already exists, what we should be doing is removing it, so this leaves things as they are until the user clicks again. This should only happen if the clicks are too close together anyway.

components/bookmark/index.jsx
ajacksified: I disagree on both counts. I don't see a problem with exporting for tests, and I don't agree that component file should only ever export a single function.

ajacksified: Disagree.

megoth: I think there are ways we can make the code more logically modular than these functions, e.g. moving domain/model logic into helpers, like we've done in `src/solidClientHelpers` (I think we could probably improve these concepts even more).

I'm not opposed to these functions; they have their place, and do them if it's the best way to get something done. But I prefer to avoid them, in preference of other patterns.

ajacksified: I only agree if we're re-using these functions; otherwise, why is moving the exported function from being exported from File A, to being exported from File B, better than exporting the functions from file A?

megoth: They must be designed with reusability in mind, of course. No point in modules for modules sake.

If there really is no other ways that makes sense than to bundle some logic that allows you to test the code easier, sure, export them from the file itself.

megoth (approved):
I don't think we need to continue this discussion of whether the logic should remain internal or exposed as methods for testability. I agree they have their uses, so I'll approve this PR.

components/bookmark/index.jsx
ajacksified: > They must be designed with reusability in mind, of course. No point in modules for modules sake.

Reusability isn't the only reason to create small modules; it also makes the code easy to read when it's small.

> If there really is no other ways that makes sense than to bundle some logic that allows you to test the code easier, sure, export them from the file itself.

I'm still curious about why it's worse to export small functions by default. So far it only seems to have upsides, and no downsides.

megoth: I don't know if a function that takes an object with ten properties can be called a small function?

ajacksified (approved):

PR #98, opened 2020-09-29
Improving the design of Navigator Context menu - especially wrt Sharing features

megoth:
# New feature description

- NB: This rips out some of the work on default permissions, as they are wildly different in the designs - most of these functions were similar to the default access functionality; we want a bit more sophisticated handling of default permissions. For now we'll manage default permission _together_ with specific permissions, meaning
  - If you're managing access on a container, altering access will change _both_ specific and default access in the ACL resource for that container
  - If you're managing access on a resource, it will only affect the specific access in the ACL resource for that container
  - If you're adding access to a resource that does not already have a specific ACL resource, it will copy the access from the fallback ACL resource, and create a specific ACL resource for you
- This also removes a lot of wiring between components by using the `DatasetProvider` from Solid React UI. Do note that it _must_ be supplied `datasetWithAcl`, otherwise it will fail (i.e. do `getSolidDatasetWithACl` instead of `getSolidDataset`).
- AlertContext is expanded - the property confirmed have now three states: `true`, `false`, and `null`. The latter is the default when no active action is chosen by the user.

# Checklist


ajacksified:
One major difference I noticed - when selecting different resources, the state of the context menu doesn't revert to "loading", it just shows the old data until the new data loads in. I think we need to show some kind of update in state - such as the loading UI.

Also, "Fetching data" looks a little odd; it doesn't have any margin or padding. Maybe use the loading spinner instead?

megoth:
> One major difference I noticed - when selecting different resources, the state of the context menu doesn't revert to "loading", it just shows the old data until the new data loads in. I think we need to show some kind of update in state - such as the loading UI.
> 
> Also, "Fetching data" looks a little odd; it doesn't have any margin or padding. Maybe use the loading spinner instead?

Ouch, ripped out a bit to much wrt the loading process. Apologies.

Just to double-check: Are you ok with ripping out the loading-components (the ones that uses Skeleton components to look similar to the ones they're loading for)? They're nice, but I think it's a bit to early for these optimizations, since some parts might change later. Also, I would like to have a consistent experience for loading, so will make use of the Spinner for the "Fetching data" part.

megoth:
Also noticed a bug that come from this PR that stops you from opening non-container resources. Will fix before next review.

megoth:
@ajacksified has fixed what you noted, and the other bug I found.

VirginiaBalseiro (approved):

ajacksified:
> Are you ok with ripping out the loading-components (the ones that uses Skeleton components to look similar to the ones they're loading for)? They're nice, but I think it's a bit to early for these optimizations, since some parts might change later. Also, I would like to have a consistent experience for loading, so will make use of the Spinner for the "Fetching data" part.

I agree a consistent experience is good, although Skeleton is useful in cases where we load extra data asynchronously - such as owner info for bookmarks. We can allow bookmarks to display their names while later on loading in owner info.

components/resourceDetails/resourceSharing/agentAccessList/index.jsx
```diff
@@ -19,109 +19,67 @@
...
 
-    const response = await saveFn({
-      iri,
-      webId,
-      access,
-      fetch,
+  useEffect(() => {
+    if (!dataset) return;
+    setLoading(true);
+    getPermissions(dataset, fetch).then((normalizedPermissions) => {
```
ajacksified, line 45: Rather than using `getPermissions` directly, you should create a hook similar to the ones in https://github.com/inrupt/pod-browser/blob/master/src/hooks/solidClient/index.js and use `swr` to improve loading times.

ajacksified (changes requested):

components/resourceDetails/resourceSharing/agentAccessList/index.jsx
megoth: I'm having trouble extracting this logic into a hook =/

Specifically because of the if-forks in https://github.com/inrupt/pod-browser/blob/design/sharing-details/src/solidClientHelpers/permissions.js#L190-L198, and into the function that fetches the profile (https://github.com/inrupt/pod-browser/blob/design/sharing-details/src/solidClientHelpers/profile.js#L52-L61).

ajacksified (approved):

PR #100, opened 2020-10-01
PodNavigator component

megoth:
# Extended PodIndicator component

Allows user to navigate to other Pods via the PodIndicator.

- Accessible containers shown in the Navigator
- Accessible resources opens the container, and the specific resource in the context menu
- Non accessible resources (401, 403) opens in Navigator with warning that the resource cannot be accessed
- Not found resources (404) open in Navigator with warning that the resource cannot be found
- Resources that cannot be fetched (e.g. CORS errors, e.g. https://google.com) are redirected to /access-required

The latter scenario should perhaps be handled better.

# Checklist


VirginiaBalseiro (approved):

ajacksified:
I seem to get 401s for containers that I should have access to:

Steps:
1. Log in with ajacksified.inrupt.net
1. Enter https://khoward.inrupt.net into the pod navigator
2. See 401 (probably accurate)
3. Enter https://ajacksified.inrupt.net to go back to my own pod
4. See 401 (inaccurate)
5. Click inrupt logo
6. View my pod's contents

Not sure why - it is sending what looks like correct dpop headers.

VirginiaBalseiro:
@ajacksified that might be related to another bug we have - it seems once you get a 401 you get "kicked out" and get 401s for all subsequent requests.

megoth:
> I seem to get 401s for containers that I should have access to:
> 
> Steps:
> 
>     1. Log in with ajacksified.inrupt.net
> 
>     2. Enter https://khoward.inrupt.net into the pod navigator
> 
>     3. See 401 (probably accurate)
> 
>     4. Enter https://ajacksified.inrupt.net to go back to my own pod
> 
>     5. See 401 (inaccurate)
> 
>     6. Click inrupt logo
> 
>     7. View my pod's contents
> 
> 
> Not sure why - it is sending what looks like correct dpop headers.

As an authenticated user, you should get `403`, not `401`. I'm not able to reproduce this bug. But isn't logging in with WebIDs on NSS servers problematic nowadays anyway? I think maybe the bug might be related to the auth library.

Are you able to reproduce this when authenticating using the broker?

I did  Maybe we should create a separate ticket for it, to get the rest of the work in this PR through?

ajacksified:
> But isn't logging in with WebIDs on NSS servers problematic nowadays anyway?

Not since we reverted to an earlier version of auth.

> Maybe we should create a separate ticket for it, to get the rest of the work in this PR through?

Because this is automatically deployed, I'd rather fix it here than push something broken.

This breaks if users don't remember to add a trailing slash - we should do that automatically. This may have been my earlier problem, as I can seem to go between NSS pods now. 

If I go to a resource I don't have access to, then try to go back to a resource I do have access too, it flashes "You do not have access" for a moment until it loads the resource I do have access to.

It also makes a HEAD to `https://ldp.demo-ess.inrupt.com/` and a GET to `https://ldp.demo-ess.inrupt.com/profile/card` when I try to go from ESS to NSS.

It should clear the textbox after I've successfully navigated to a new pod.

megoth:
> This breaks if users don't remember to add a trailing slash - we should do that automatically. This may have been my earlier problem, as I can seem to go between NSS pods now.

Fixed.

> If I go to a resource I don't have access to, then try to go back to a resource I do have access too, it flashes "You do not have access" for a moment until it loads the resource I do have access to.

Think this is fixed. There was a bug caused by Solid Client because even though I'd made sure to append a slash at the end, Solid Client would make another request where the slash was removed. I haven't been able to reproduce this after some changes, so I think it's fixed.

> It also makes a HEAD to `https://ldp.demo-ess.inrupt.com/` and a GET to `https://ldp.demo-ess.inrupt.com/profile/card` when I try to go from ESS to NSS. (My guess is ESS -> NSS won't ever work anyways, because the token format is different, but we should handle the error gracefully.)

Should be fixed.

> It should clear the textbox after I've successfully navigated to a new pod.

Fixed.

> It should show a useful error if you attempt to go to a location which is not a container, rather than a blank screen (for example, google.com) - preferably it would attempt a HEAD on the resource and display an error on anything other than a 200 response for a container-type resource.

All URLs will now be shown in the Navigator, even those that shouldn't, either with a **Access Forbidden** (401, 403), **Not Found** (404), or **Not Supported** (non-containers, all other errors, e.g. CORS, 500).

ajacksified:
Typing in "asfasdf" for a URL gives me a blank screen rather than an error; can we use html5 form field validation with type="url" here?

components/header/podIndicator/index.jsx
```diff
@@ -19,55 +19,97 @@
...
+export const closeHandler = (setAnchorEl) => () => setAnchorEl(null);
+
+export const submitHandler = (handleClose, setUrl) => async (
+  event,
+  url,
+  router
+) => {
+  event.preventDefault();
+  const containerUrl = normalizeContainerUrl(url);
+  await router.replace("/resource/[iri]", resourceHref(containerUrl));
```
ajacksified, line 49: This should be `push` so that I can go "back" in my browser history.

components/header/podIndicator/index.jsx
```diff
@@ -19,55 +19,97 @@
...
+          }}
         >
-          {podOwnerName}
-        </button>
+          <Form onSubmit={(event) => onSubmit(event, url, router)}>
+            <Input
+              id="PodNavigator"
+              label="Go to Pod"
+              value={url}
+              onChange={(event) => setUrl(event.target.value)}
```
ajacksified, line 109: Add `type="url"` to get validation for free.

megoth: Didn't use required (it sets the input to `:invalid` from the start), but checks it as part of submitting form.

components/header/podIndicator/index.jsx
megoth: Fixed

ajacksified (approved):

ajacksified (approved):

PR #101, opened 2020-10-05
Fix/duplicate contacts webids

VirginiaBalseiro:
### Disallowing duplicate WebIDs to be added to contacts
- If the user tries to add a contact with a WebId that is already in the address book, an error will be displayed and the contact won't be added
- Added tests to the AddContact component

### DIFF
--- a/components/addContact/index.jsx
+++ b/components/addContact/index.jsx
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { useContext } from "react";
+import React, { useContext, useState } from "react";
 import { createStyles, makeStyles } from "@material-ui/styles";
 import { useBem } from "@solid/lit-prism-patterns";
 import clsx from "clsx";
@@ -28,7 +28,7 @@ import { getStringNoLocale, getUrl } from "@inrupt/solid-client";
 import { vcard, foaf, space } from "rdf-namespaces";
 import { useSession } from "@inrupt/solid-ui-react";
 import Spinner from "../spinner";
-import { saveContact } from "../../src/addressBook";
+import { findContactInAddressBook, saveContact } from "../../src/addressBook";
 import { getResource } from "../../src/solidClientHelpers/resource";
 import { joinPath } from "../../src/stringHelpers";
 import AgentSearchForm from "../agentSearchForm";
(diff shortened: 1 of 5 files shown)


components/addContact/index.jsx
```diff
@@ -38,11 +38,33 @@ import { useRedirectIfLoggedOut } from "../../src/effects/auth";
...
+  alertError,
+  alertSuccess,
+  fetch,
+  webId,
+}) {
   return async (iri) => {
+    setIsLoading(true);
     const { response: userProfile } = await getResource(webId, fetch);
     const pod = getUrl(userProfile.dataset, space.storage);
+    const addressBookIri = joinPath(pod, "contacts/");
```
megoth, line 56: Use the hook `useAddressBook` instead, and use `getSourceUrl` from the Solid Client JS to get the url.

components/addContact/index.jsx
```diff
@@ -52,14 +74,10 @@ export function handleSubmit({ alertError, alertSuccess, fetch, webId }) {
     if (profileError) alertError(profileError);
     if (profile) {
       const contact = { webId: iri };
-      const name = getStringNoLocale(profile.dataset, foaf.name);
-      const fn = getStringNoLocale(profile.dataset, vcard.fn);
-
-      if (name) contact.fn = name;
-      if (fn) contact.fn = fn;
-
+      contact.fn =
```
megoth, line 77: Maybe lets use `fetchProfile` instead, so we have a "standard" way of handling names, in a prepared object.

components/addContact/index.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+    const alertError = jest.fn();
+    const alertSuccess = jest.fn();
+    const handler = handleSubmit({
+      setIsLoading,
+      alertError,
+      alertSuccess,
+      fetch,
+      webId: personUri,
+    });
+    getUrl.mockReturnValue("https://www.example.com/");
```
megoth, line 117: You shouldn't need to mock the return value here if you've used the mocking methods from the Solid Client JS.

components/addContact/index.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      alertError,
+      alertSuccess,
+      fetch,
+      webId: personUri,
+    });
+    getResource.mockResolvedValue({
+      response: { dataset: personDataset, iri: personUri },
+      error: null,
+    });
+    getUrl.mockReturnValue("https://www.example.com/");
```
megoth, line 157: Again, these methods that don't do any fetch shouldn't be necessary, use the mocking methods from Solid Client properly instead.

components/addContact/index.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      mockThingFrom,
+      addStringNoLocale,
+    } = jest.requireActual("@inrupt/solid-client");
+    const personUri = "http://example.com/alice#me";
+    const personDataset = addStringNoLocale(
+      mockThingFrom(personUri),
+      vcard.fn,
+      "Alice"
+    );
+    const peopleDataset = mockSolidDatasetFrom(
```
megoth, line 139: Use `contactsContainerIri("https://www.example.com")` instead of hardcoding       `"https://www.example.com/contacts"`.

megoth (changes requested):

package-lock.json
```diff
@@ -5143,17 +5143,6 @@
         "@rdfjs/data-model": "^1.1.1"
       }
     },
-    "@sentry/browser": {
```
megoth, line 5146: Seems weird to have these changes in package-lock.json, when there are no changes in package.json. Sure this is right?

megoth (changes requested):

megoth (approved):

ajacksified (changes requested):
I was able to add a duplicate. Can dig through the code to figure out why in a few - just want to block the approval.

Edit: tried again and it's fine.

ajacksified (approved):

PR #114, opened 2020-10-07
Make saveContact generic to type

ajacksified:
This modifies `saveContact` to map types to container / file names.

We are currently storing all person contacts in “people.ttl” and the “People/” container. I have updated the code to allow for switching the index file and container name based on type of contact - which is currently pulled from the type of the profile (reading the Thing by foaf.primaryTopic or foaf.maker when defined.) If the profile is a foaf.Person (regardless of it it has multiple types), we treat it as a contact and add it to “people”, or otherwise we display an error message.

Using a specific file name (people.ttl, People/) is pretty hacky compared to including the contact type in an overall index file, but this maintains compatibility with databrowser.

### DIFF
--- a/__testUtils/mockContactResource.js
+++ b/__testUtils/mockContactResource.js
@@ -19,29 +19,16 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import useSWR from "swr";
-import { useSession } from "@inrupt/solid-ui-react";
-import { getSourceUrl } from "@inrupt/solid-client";
-import { foaf } from "rdf-namespaces";
-import { getContacts } from "../../addressBook";
+import { addUrl, mockThingFrom } from "@inrupt/solid-client";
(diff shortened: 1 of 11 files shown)


megoth (approved):
This works given that `foaf:maker` (or `foaf:primaryTopic`) for the profile document points to the contact we want to add. I think this is always the case for profiles on NSS and ESS, but I'm not 100% sure this will keep up in the future. (E.g. some might say for an organization that the co-founder(s) of the organization should be referred to as the maker of the profile document for that organization.) There's also more custom profiles, like Sarven's (https://csarven.ca/#i) that doesn't seem to use this predicate.

That being said, I think the assumption is ok, just wanted to mention this so we're aware of it in the future.

ajacksified:
> This works given that `foaf:maker` (or `foaf:primaryTopic`) for the profile document points to the contact we want to add.

I pulled this from the spec.

VirginiaBalseiro (approved):

kevin-inrupt (approved):

megoth (approved):

PR #120, opened 2020-10-09
Feature/delete contact

andydavison:
# New feature description
- Adds contact context menu drawer
- Contact table rows are selectable; selecting a contact opens context drawer
- Adds contact delete functionality - removes relevant index.ttl and containing folder
- Splits DeleteLink into DeleteLink and DeleteResourceLink, to allow reuse of confirmation/alert handler logic by DeleteContactLink
- Refactors getPeople/usePeople to return people contacts as stored in address book
- Adds getProfiles/useProfiles to return profiles of given people contacts (as originally returned by returned by getPeople)

Still to add:
- Currently fixing new tests for contact delete (not yet pushed)
- Possible refactoring to account for fixed people.ttl, removing Person related hardcoding


andydavison:
Removing the draft status as the functionality and the pushed tests are ready to be looked at, but there's a few more tests to come shortly

megoth (approved):

megoth (changes requested):
Sorry, misclick earlier - I haven't reviewed it properly yet, will start right away. I'll change back to approve once I've reviewed it properly.

components/deleteContactLink/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+import React from "react";
+import T from "prop-types";
+import DeleteLink from "../deleteLink";
+
+/* eslint react/jsx-props-no-spreading: 0 */
+export default function DeleteContactLink({ onDelete, ...linkProps }) {
+  return (
+    <DeleteLink
+      confirmationTitle="Delete Contact?"
```
ajacksified, line 30: So that it's similar to deleting a file, the title should be "Delete Contact" and the body should include the name: `Are you sure you wish to delete ${name} from your contacts?`

src/addressBook/index.js
```diff
@@ -342,3 +349,14 @@ export async function saveContact(addressBookIri, schema, fetch) {
...
+
+export async function deleteContact(contactToDelete, fetch) {
+  const contactIri = contactToDelete.iri;
+  const contactContainerIri = contactIri.substring(
+    0,
+    contactIri.lastIndexOf("/") + 1
+  );
+
+  await deleteFile(contactIri, { fetch });
+  await deleteFile(contactContainerIri, { fetch });
```
ajacksified, line 361: Could these be run in parallel with Promise.all?

ajacksified: The link to the contact also needs to be removed from `people.ttl`.

ajacksified (changes requested):

components/deleteContactLink/index.jsx
andydavison: cool cool, was going off the designs for the copy — will update 👍

src/addressBook/index.js
megoth: I think the latter will fail if for some reason it is resolved by the server before the former. So I think they need to be serialized.

The request to delete it from the People Index can be done parallell with any of them though.

andydavison: deleting a non-empty folder throws an error, so unfortunately no it seems.

and yep, until people.ttl was fixed we couldn't actually identify the entry to remove — can do this now though

components/contactsList/index.jsx
```diff
@@ -54,19 +57,48 @@ function ContactsList() {
...
+      onDelete={handleDeleteContact}
+    />
+  );
 
-  if (isLoading) return <Spinner />;
   if (addressBookError) return addressBookError;
   if (peopleError) return peopleError;
 
+  const isLoading =
+    (!addressBook && !addressBookError) ||
```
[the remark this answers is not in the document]
andydavison: true — didn't write the condition, was just switching the order of those lines and didn't amend it (previously it could return the spinner even if there was error). Will fix!

src/addressBook/index.js
andydavison: > The link to the contact also needs to be removed from `people.ttl`.

I think that bit will be easier to do after/along with https://inrupt.atlassian.net/browse/SOLIDOS-503. This does now update people.ttl though!

ajacksified:
I'm getting `TypeError: Cannot read property 'internal_localSubject' of undefined` when I attempt to delete a contact.

VirginiaBalseiro:
> I'm getting `TypeError: Cannot read property 'internal_localSubject' of undefined` when I attempt to delete a contact.

@ajacksified can you try this again? I could not reproduce this.

ajacksified:
Looks good after I completely cleared out my contacts - may just have had some bad data in there.

ajacksified (approved):

megoth (approved):

PR #121, opened 2020-10-09
Update default domain lists

ajacksified:


kevin-inrupt (approved):
https://inrupt.net is the only option displayed in the Login combobox.

PR #127, opened 2020-10-13
Fix: variable header height in Firefox

VirginiaBalseiro:
This PR fixes height being ignored in Firefox when using flexbox, by setting the `div` container inside the header to a fixed height and thus forcing the height to be 100px.

This issue seems to be due to the app layout container being a flex container, rather than the header container.

### DIFF
--- a/components/header/styles.ts
+++ b/components/header/styles.ts
@@ -21,14 +21,21 @@
 
 import { PrismTheme, createStyles } from "@solid/lit-prism-patterns";
 
+const HEADER_HEIGHT = "100px";
+
 const styles = (theme: PrismTheme) =>
   createStyles(theme, ["appLayout", "headerBanner"], {
     logoIndicatorContainer: {
       display: "flex",
+      height: HEADER_HEIGHT,
       flexDirection: "column",
       alignItems: "center",
+      justifyContent: "center",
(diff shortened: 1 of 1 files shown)


components/header/styles.ts
```diff
@@ -21,14 +21,21 @@
...
   createStyles(theme, ["appLayout", "headerBanner"], {
     logoIndicatorContainer: {
       display: "flex",
+      height: HEADER_HEIGHT,
       flexDirection: "column",
       alignItems: "center",
+      justifyContent: "center",
       marginRight: theme.spacing(2),
-      paddingTop: theme.spacing(2),
+      paddingTop: theme.spacing(1),
```
megoth, line 35: Why the need for paddingTop? This increases the total height to be over 100px (either 120px or 140px).

I would've hoped for a solution to be implemented in Prism? Any reason why you didn't prefer that?

megoth (changes requested):

components/header/styles.ts
VirginiaBalseiro: I just put a PR for another issue in Prism that might fix this so turning this into a draft

VirginiaBalseiro:
Closed in favor of https://github.com/inrupt/prism/pull/105

PR #129, opened 2020-10-14
Fix/agent search form validation

VirginiaBalseiro:
This PR fixes:
- Uncaught promise if "Add Contact" button clicked without WebId
- "Insecure Resource Requested" when trying to add a WebId with `http`

## To test
- This component can be tested in Add Contact and in Permissions for an individual resource
- Verify that the form cannot be submitted without a valid url that doesn't start with `https`

### DIFF
--- a/components/agentSearchForm/index.jsx
+++ b/components/agentSearchForm/index.jsx
@@ -22,34 +22,39 @@
 import React, { useState } from "react";
 import { useId } from "react-id-generator";
 import T from "prop-types";
-import { Button, Input } from "@inrupt/prism-react-components";
-
-export function handleClick({ setAgentId, onSubmit }) {
-  return (agentId) => {
-    setAgentId("");
-    onSubmit(agentId);
-  };
-}
-
-export function handleChange(setAgentId) {
-  return ({ target: { value } }) => {
-    setAgentId(value);
-  };
-}
+import { Form, Button, Input } from "@inrupt/prism-react-components";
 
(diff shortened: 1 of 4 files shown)


components/agentSearchForm/index.jsx
```diff
@@ -22,34 +22,38 @@
...
   return (
-    <>
-      <Input id={inputId} label="WebID" onChange={onChange} value={agentId} />
+    <Form onSubmit={handleSubmit}>
+      <Input
+        id={inputId}
+        label="WebID"
+        onChange={handleChange}
+        value={agentId}
+        required
```
ajacksified, line 47: Arne noted in #100 that `required` gave the textbox a red border when you load the page- we may want to remove this field and check it in onSubmit instead.

megoth, line 47: This makes the field invalid from the getgo - Although not optimal, I think we should not use this and rather have a test in the submit function that checks if there's a value before submitting (i.e. [what I did in this PR](https://github.com/inrupt/pod-browser/pull/100/commits/d0aa1b6135fd3c9ce21bc1bd48e79e01329d5f89#diff-1f62fb90f81f3d9cdb374658a11dc25e234a621460729589a97797d315f1fcccR51-R53)).

megoth (changes requested):

megoth (changes requested):
Given the custom validation of empty strings there should also be a test that verifies that the form is not submitted for empty strings.

ajacksified (approved):

PR #130, opened 2020-10-14
Implemented access control strategies for WAC and ACP

megoth:
This PR restructures the existing WAC-specific code into a strategy. It also creates another strategy for ACP-specific code. These two can be be used interchangeable, and should allow us to use the existing access management UX on servers that support WAC or ACP.

Note that the ACP-specific stuff uses mocked API, since the ACP API in Solid Client JS is not complete.

# Testing

You cannot test the ACP specific code yet. Testing this should therefore be to check that there are no regression bugs.

# Checklist


src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+      .flat();
+    (await this.getPolicies(policyUrls)).forEach(({ modes, agents }) =>
+      agents.forEach((webId) => {
+        const permission = getOrCreatePermission(permissions, webId);
+        permission.acp.apply = addAcpModes(permission.acp.apply, modes);
+        permissions[webId] = permission;
+      })
+    );
+    // assigning Control
+    const controlPolicyUrls = getControlPolicyUrlAll(
```
Vinnl, line 180: I think this will be called `getReferencedPolicyUrlAll`.

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+      this.#datasetWithAcr
+    ).filter((url) => url === this.#policyUrl);
```
Vinnl, line 182: Will this not simply lead to

```
const controlPolicyUrls = [this.#policyUrl]
```

?

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+    const accessControls = getAccessControlAll(this.#datasetWithAcr);
+    let accessControlWithPolicy = accessControls.find((ac) =>
+      getPolicyUrlAll(ac).find(policyUrl)
+    );
+    if (!accessControlWithPolicy) {
+      accessControlWithPolicy = chain(createAccessControl(), (ac) =>
+        addPolicyUrl(ac, policyUrl)
+      );
+    }
+    this.#datasetWithAcr = setAccessControl(
```
Vinnl, line 221: Hmm, so this updates `this.#datasetWithAcr` to add the Access Control, but it doesn't actually save it. Which makes me think - I don't think we actually have an API for saving ACRs yet? I've just added a ticket to do so, with the proposed name `saveAccessControlResource`, which would take `this.#datasetWithAcr` as a parameter.

Vinnl: Ah wait, I see you're calling a regular `saveSolidDatasetAt()` below, which will make sense if we ever get atomic updates. Until then, though, I think we'll keep them as separate updates like we do for ACLs, to ensure that as a developer, you can deal with either the SolidDataset- or the ACR-update request failing.

Vinnl (commented):
Mainly looked at the ACP stuff. Added some general thoughts that I just couldn't resist writing down (but feel free to ignore), and some more relevant points about the solid-client API for ACPs. Found at least one extra ticket of work that's needed on the solid-client side, so that's great :)

components/resourceDetails/resourceSharing/agentAccessList/index.test.jsx
```diff
@@ -40,14 +39,9 @@ const permission = {
...
-  let SessionProvider;
-
-  beforeEach(() => {
-    session = mockSession();
-    SessionProvider = mockSessionContextProvider(session);
-    jest.spyOn(permissionHelpers, "getPermissions").mockResolvedValue([]);
-  });
+  const accessControl = {
+    getPermissions: () => Promise.resolve([]),
+  };
```
andydavison, line 44: can this just use the new mockAccessControl?

components/resourceDetails/resourceSharing/agentAccessList/index.test.jsx
```diff
@@ -56,24 +50,24 @@ describe("AgentAccessList", () => {
...
           </DatasetProvider>
-        </SessionProvider>
+        </AccessControlProvider>
       )
     ).toMatchSnapshot();
   });
 
   it("renders a list of permissions", () => {
-    permissionHelpers.getPermissions.mockResolvedValue([permission]);
+    accessControl.getPermissions = () => Promise.resolve([permission]);
```
andydavison, line 63: can we do this without mutating the shared `accessControl`? Works fine for now, but if we add subsequent tests then they'll be affected by the order tests run in

components/resourceDrawer/index.jsx
```diff
@@ -42,39 +44,42 @@ export function handleCloseDrawer({ setMenuOpen, router }) {
...
-    const encodedResourceIri = encodeURI(resourceIri);
-    getResourceInfoWithAcl(encodedResourceIri, { fetch }).then((dataset) => {
-      setDatasetWithAcl(dataset);
-      setLoading(false);
-    });
-  }, [fetch, resourceIri]);
+    if (accessControlError || resourceError) {
+      throw accessControlError || resourceError;
+    }
+  });
```
andydavison, line 68: does this need to be in a useEffect anymore?

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+} from "../../solidClientHelpers/utils";
+import { getOrCreateDataset } from "../../solidClientHelpers/resource";
+
+const POLICIES_CONTAINER = "policies/";
+
+export function getPoliciesContainerUrl(podRootUri) {
+  return joinPath(podRootUri, POLICIES_CONTAINER);
+}
+
+export function createAcpMap(read = false, write = false, append = false) {
```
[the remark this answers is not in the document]
megoth: Fair point - I simply followed the convention that was established in PB earlier (Read, Write, Append, Control), which is what is presented in the access management UI (and you can see other examples of this in `src/solidClientHelpers/permissions.js`). I'm not sure why it was set in that order though, especially since [newer designs](https://www.figma.com/file/W2o8rM7yu5WxuRAlAOukuG/Pod-Browser?node-id=650%3A77304) seems to present them in the order you propose (Read, Append, Write, Control).

I don't want to change to much now though, so I'll leave this as is for now.

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+    acp.apply.append,
+    acp.access.read && acp.access.write
+  );
+}
+
+export function getOrCreatePermission(permissions, webId) {
+  const permission = permissions[webId] || {
+    webId,
+  };
+  permission.acp = permissions.acp || {
```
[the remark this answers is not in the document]
megoth: As you say, shouldn't really matter in this case, but I agree with the use of `??` over `||` in general.

src/solidClientHelpers/utils.js
```diff
@@ -230,3 +230,14 @@ export const ERROR_CODES = {
 export function isHTTPError(errorMessage, code) {
   return !!errorMessage.match(new RegExp(code));
 }
+
+export function sharedStart(...strings) {
```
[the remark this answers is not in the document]
megoth: It returns the shared characters that all strings starts with, e.g. `sharedStart("bar", "baz", "bam")` will return "ba". I'm going to add a comment about this, since it's not apparent.

I needed to use it for `getPolicyUrl` where I want to find the corresponding path for the policy resource for a given resource. I compare the URI for the resource with the URI policies container, and generate a path based on this comparison.

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+        getSourceUrl(updatedDataset),
+        updatedDataset
+      );
+    } catch (err) {
+      return error(err.message);
+    }
+    return respond(this.#datasetWithAcr);
+  }
+
+  static async init(resourceInfo, policies, fetch) {
```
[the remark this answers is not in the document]
megoth: The policies will be stored in a policies container that's located on the root (e.g. `https://megoth-demo2.inrupt.net/policies/`). Renaming it `policiesContainer` will be clearer yes, so I'll do that.

src/accessControl/acp/index.js
megoth: As I understand it this will lead to either `[this.#policyUrl]` or `[]`, given whether there is a policy set for control or not.

src/accessControl/acp/index.js
megoth: Renamed

src/accessControl/acp/index.js
megoth: I've introduced `saveAccessControlResource`, I'll push it with my other updates very soon.

components/resourceDetails/resourceSharing/agentAccessList/index.test.jsx
megoth: Good point, I've changed this so that each test have their own unique instance of a mocked access control.

src/accessControl/acp/index.js
Vinnl: OK. Just noted it because it looked weird to me to create an array with just either one or zero elements, and then `forEach`'ing over that, but I suppose that makes it easier with `this.getPolicies` :)

components/resourceDrawer/index.jsx
megoth: Fair point, it doesn't need to no. We should also handle the error better. I'll create a ticket to handle this, since the error handling can be done in various ways.

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+  const newPolicy = createPolicy(url);
+  const newDataset = setPolicy(dataset, newPolicy);
+  return { policy: newPolicy, dataset: newDataset };
+}
+
+export async function setAgents(policy, webId, access, mode) {
+  const { agents: existingAgents } = getRequiredRuleOnPolicy(policy);
+  const agentIndex = existingAgents.indexOf(webId);
+  let agents;
+  if (access[mode] && agentIndex === -1) {
```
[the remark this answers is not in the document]
megoth: Yes, this checks whether a given mode is true (i.e. whether Read was checked in the access management UI) _and_ whether the agent does not already exist in the list. If so, it adds the agent to the list. If the mode is false _and_ the agent exists in the list, it is removed.

I thought by setting the policy names to specific instances (`<policyResource>#readPolicy`, `<policyResource>#writePolicy`, etc) we would avoid these problems with agents? Of course, if people start manually editing the policy resource, and change how the various policies work, we're gonna have problems. But I would argue that's outside the scope of error handling we can expect to handle with this setup?

src/accessControl/acp/index.js
```diff
@@ -0,0 +1,305 @@
...
+  if (existingPolicy) {
+    return { policy: existingPolicy, dataset };
+  }
+  const newPolicy = createPolicy(url);
+  const newDataset = setPolicy(dataset, newPolicy);
+  return { policy: newPolicy, dataset: newDataset };
+}
+
+export async function setAgents(policy, webId, access, mode) {
+  const { agents: existingAgents } = getRequiredRuleOnPolicy(policy);
```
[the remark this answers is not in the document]
megoth: So the object that I was expecting from `getRequiredRuleOnPolicy` is not the case?

```
type AccessRule = {
  agents: WebId[];
  groups: UrlString[];
  creator: boolean;
  public: boolean;
  authenticatedAgent: boolean;
}
```

I have to get that data using other methods?

src/accessControl/acp/index.js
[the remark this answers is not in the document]
Vinnl: Yeah sorry, I hadn't gotten to that part of the code yet when I saw this. Yes, it should only break if the Policies you create are changed.

src/accessControl/acp/index.js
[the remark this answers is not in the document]
Vinnl: Yeah exactly, instead you'll have functions like `hasPublicInRule`, which you'll have to pass a Rule, rather than a Policy.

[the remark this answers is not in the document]
megoth: I think I've been able to address this in the latest couple of commits.

andydavison (approved):

src/accessControl/acp/mockedClientApi.js
```diff
@@ -49,6 +49,10 @@ export function getRequiredRuleOnPolicyAll(policy) {
   throw new Error("Not implemented");
 }
 
+export async function getSolidDatasetWithAcp(url, options) {
```
Vinnl, line 52: Small heads-up: this will be called `getSolidDatasetWithAcr` I think.

megoth: Noted - will fix that in later PRs when we're connecting to the actual APIs.

Vinnl (approved):

ajacksified (approved):

PR #133, opened 2020-10-16
Feat: add pod ownership to useOwnerProfile hook

VirginiaBalseiro:
# Add pod ownership to useOwnerProfile hook
This enables the pod indicator to use pod owner when available, or fall back to fetching the profile from the constructed profile iri. 

## To test
- Log in with an ESS and check that the owner name is displayed properly
- Do the same with an NSS account to check that the fallback works

# Checklist

### DIFF
--- a/src/hooks/usePodOwner/index.js
+++ b/src/hooks/usePodOwner/index.js
@@ -22,24 +22,32 @@
 import { useState, useEffect } from "react";
 import { getResourceInfo, getPodOwner } from "@inrupt/solid-client";
 import { useSession } from "@inrupt/solid-ui-react";
+import { joinPath } from "../../stringHelpers";
+import usePodRoot from "../usePodRoot";
 
-export default function usePodOwner(resourceIri) {
+export default function usePodOwner({ resourceIri }) {
   const { fetch } = useSession();
   const [podOwnerWebId, setPodOwnerWebId] = useState(null);
-  const [error, setError] = useState();
+  const [error, setError] = useState(null);
+  const decodedResourceUri = decodeURIComponent(resourceIri);
+  const podRoot = usePodRoot(decodedResourceUri, null);
+  const profileIri = podRoot && joinPath(podRoot, "profile/card#me");
 
   useEffect(() => {
-    if (!resourceIri) return;
+    if (!resourceIri) {
+      setError(null);
+      setPodOwnerWebId(null);
+      return;
+    }
     (async () => {
       try {
         const resourceInfo = await getResourceInfo(resourceIri, { fetch });
-        const { response: webId } = await getPodOwner(resourceInfo, fetch);
+        const { response: webId } = getPodOwner(resourceInfo, fetch);
         setPodOwnerWebId(webId);
       } catch (e) {
         setError(e);
       }
     })();
   }, [resourceIri, fetch]);
-
-  return { podOwnerWebId, error };
+  return { podOwnerWebId: podOwnerWebId || profileIri, error };
 }
--- a/src/hooks/usePodOwner/index.js
+++ b/src/hooks/usePodOwner/index.js
@@ -23,14 +23,13 @@ import { useState, useEffect } from "react";
 import { getResourceInfo, getPodOwner } from "@inrupt/solid-client";
 import { useSession } from "@inrupt/solid-ui-react";
 import { joinPath } from "../../stringHelpers";
-import usePodRoot from "../usePodRoot";
+import usePodRootUri from "../usePodRootUri";
 
 export default function usePodOwner({ resourceIri }) {
   const { fetch } = useSession();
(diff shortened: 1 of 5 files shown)


src/hooks/usePodOwnerProfile/index.js
```diff
@@ -21,21 +21,40 @@
...
   const decodedResourceUri = decodeURIComponent(router.query.iri);
+  const [podOwnerWebId, setPodOwnerWebId] = useState(null);
   const podRoot = usePodRoot(decodedResourceUri, null);
-  const profileIri = podRoot && joinPath(podRoot, "profile/card#me"); // we won't need to do this once ownership is available
+  const profileIri =
+    podOwnerWebId || (podRoot && joinPath(podRoot, "profile/card#me")); // we won't need to do this once ownership is available
   const { data: authProfile, error: authError } = useAuthenticatedProfile();
   const { data: ownerProfile, error: ownerError } = useFetchProfile(profileIri);
 
+  useEffect(() => {
```
megoth, line 45: Let's create this into a hook instead, makes it easier to reuse the functionality.

megoth (changes requested):

src/hooks/usePodOwnerProfile/index.js
```diff
@@ -25,14 +25,17 @@ import { joinPath } from "../../stringHelpers";
...
 import usePodRoot from "../usePodRoot";
 import useFetchProfile from "../useFetchProfile";
+import usePodOwner from "../usePodOwner";
 
 export default function usePodOwnerProfile() {
   const [profile, setProfile] = useState();
   const [error, setError] = useState();
   const router = useRouter();
   const decodedResourceUri = decodeURIComponent(router.query.iri);
+  const { podOwnerWebId } = usePodOwner(router.query.iri);
```
ajacksified, line 35: This will return null while the request is being made, causing the first load to attempt to load profile/card#me. You may need to return a `requestInProgress` boolean so that, later on in the effect to setProfile, you only set the profile iri if the request is _not_ in progress.

ajacksified (changes requested):

src/hooks/usePodOwner/index.js
```diff
@@ -0,0 +1,45 @@
...
+import { getResourceInfo, getPodOwner } from "@inrupt/solid-client";
+import { useSession } from "@inrupt/solid-ui-react";
+
+export default function usePodOwner(resourceIri) {
+  const { fetch } = useSession();
+  const [podOwnerWebId, setPodOwnerWebId] = useState(null);
+  const [error, setError] = useState();
+
+  useEffect(() => {
+    if (!resourceIri) return;
```
megoth, line 32: You should set both `error` and `podOwnerWebId` to null in this case, as there are cycles where resourceIri will be undefined by design (e.g. the caller needs to load something before resourceIri is ready), and we don't want to use older instances of `error` and `podOwnerWebId` in those cases.

src/hooks/usePodOwnerProfile/index.js
megoth: You should also extend the tests for `userPodOwnerProfile` with a scenario that successfully uses `usePodOwner`.

megoth (changes requested):
We should have a test of `usePodOwner`. You might also want to consider pulling in the fallback (i.e. `joinPath(podRoot, "profile/card#me")`) into this hook, so that all logic on getting a podOwner WebID URL for a given resource is given in `usePodOwner` while `usePodOwnerProfile` concerns itself with returning a PodOwner in general.

megoth (approved):

ajacksified (approved):

kevin-inrupt (approved):

src/hooks/usePodOwnerProfile/index.js
```diff
@@ -25,14 +25,16 @@ import { joinPath } from "../../stringHelpers";
...
 export default function usePodOwnerProfile() {
   const [profile, setProfile] = useState();
   const [error, setError] = useState();
   const router = useRouter();
-  // const decodedResourceUri = decodeURIComponent(router.query.iri);
   const podRoot = usePodRootUri(router.query.iri, null);
-  const profileIri = podRoot && joinPath(podRoot, "profile/card#me"); // we won't need to do this once ownership is available
+  const { podOwnerWebId } = usePodOwner({ resourceIri: router.query.iri }); // passing in an object for testing purposes
+  const profileIri =
+    podOwnerWebId || (podRoot && joinPath(podRoot, "profile/card#me")); // we won't need to do this once ownership is available
```
ajacksified, line 37: With this logic, we end up trying to load `podRoot/profile/card#me` first. If you exposed a `requestInProgress` variable from `usePodOwner`, you would be able to default this to null and wait for the request to complete before loading the fallback.

ajacksified (changes requested):

ajacksified (approved):

PR #137, opened 2020-10-21
Adds onSubmit to forms that are missing it

megoth:
I've added onSubmit handling (and removed the existing onClick handling) for

- Create folder
- Managing existing access for an agent

There is a form that I haven't fixed yet, because I'm waiting for a fix in the Solid UI React. I'll create a separate PR for it later.

I also checked the following forms, which was already fixed:

- Pod Navigator
- Adding new contact

### DIFF
--- a/components/addFolderFlyout/__snapshots__/index.test.jsx.snap
+++ b/components/addFolderFlyout/__snapshots__/index.test.jsx.snap
@@ -50,6 +50,11 @@ exports[`AddFolderFlyout Renders an Add Folder button 1`] = `
               "vertical": "bottom",
             }
           }
+          classes={
+            Object {
+              "paper": "makeStyles-paper-1",
+            }
+          }
           onClose={[Function]}
           open={false}
           transformOrigin={
@@ -69,7 +74,7 @@ exports[`AddFolderFlyout Renders an Add Folder button 1`] = `
             }
             classes={
               Object {
-                "paper": "MuiPopover-paper",
+                "paper": "MuiPopover-paper makeStyles-paper-1",
                 "root": "MuiPopover-root",
               }
             }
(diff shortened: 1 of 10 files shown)


VirginiaBalseiro (changes requested):
When I try creating a new folder, the form only submits when I press enter, not when I click on the button.

megoth:
> When I try creating a new folder, the form only submits when I press enter, not when I click on the button.

Ooof, that's embarassing. But turns out that adding `type="submit"` to Material UI Button does not do the trick, and there's no apparent work-around except putting a normal button inside an `IconButton`. So I opted for switching button and input field into Prism button and input, which makes it more coherent with other forms.

VirginiaBalseiro:
> > When I try creating a new folder, the form only submits when I press enter, not when I click on the button.
> 
> Ooof, that's embarassing. But turns out that adding `type="submit"` to Material UI Button does not do the trick, and there's no apparent work-around except putting a normal button inside an `IconButton`. So I opted for switching button and input field into Prism button and input, which makes it more coherent with other forms.

Still not working for me. I thought Vercel's Preview might have not updated so I checked out the branch and ran it locally and it's still not submitting on click.

megoth:
> Still not working for me. I thought Vercel's Preview might have not updated so I checked out the branch and ran it locally and it's still not submitting on click.

Ok, this is not my day, forgot to actually change type on button in last commit... Thought I'd tested, but obviously went a bit faster than I should've. Apologies. It should be fixed now.

VirginiaBalseiro (approved):

kevin-inrupt (changes requested):
QA issues.

kevin-inrupt (approved):

PR #139, opened 2020-10-22
adding a maxWidth for small screens

VirginiaBalseiro:
This PR fixes the pod navigator going off the screen when the keyboard comes up in iOS devices.

## To test:
**Note**: needs to be tested on an actual iOS device (not browser emulators).
- Open the Vercel preview in a mobile device (iPhone)
- Open the Pod Navigator
- Start typing
- Verify that the PodNavigator is still fully  visible and usable

### DIFF
--- a/components/header/podIndicator/styles.js
+++ b/components/header/podIndicator/styles.js
@@ -38,7 +38,10 @@ const styles = (theme) =>
     },
     popover: {
       padding: theme.spacing(2, 2, 0),
-      width: 450,
+      maxWidth: "100vw",
+      [theme.breakpoints.up("sm")]: {
+        width: 450,
+      },
     },
   });
 


components/header/podIndicator/styles.js
```diff
@@ -38,7 +38,10 @@ const styles = (theme) =>
     },
     popover: {
       padding: theme.spacing(2, 2, 0),
-      width: 450,
+      maxWidth: "100vw",
```
megoth, line 41: I'm don't think the `100vw` is important here, e.g. you could've done `initial` as well. That said, it does make it so that the input field is not wider than the screen available, so I'm ok with this solution.

megoth (approved):

ajacksified (approved):

kevin-inrupt (approved):

PR #140, opened 2020-10-23
Add ellipsis cutoff to long names in podIndicator

andydavison:
- Adds max width to pod indicator name, with ellipsis cutoff
- Adds title attribute to button, with full name

Explored solutions wrapping the pod name — however, the headers in Prism are a fixed height 100px, which means there isn’t room for even a second line without shuffling things. _Could_ allow the header to grow vertically, but either it could do so indefinitely, or we’d probably want to truncate the name at some point anyway.

### DIFF
--- a/components/header/podIndicator/styles.js
+++ b/components/header/podIndicator/styles.js
@@ -31,10 +31,16 @@ const styles = (theme) =>
       fontWeight: theme.typography.fontWeightMedium,
     },
     indicatorPrompt: {
+      display: "inline-flex",
       color: theme.palette.text.secondary,
       textTransform: "none",
       fontSize: "0.825rem",
       fontWeight: theme.typography.fontWeightRegular,
+    },
+    indicatorName: {
+      textOverflow: "ellipsis",
+      overflow: "hidden",
+      maxWidth: "200px",
     },
     popover: {
       padding: theme.spacing(2, 2, 0),
--- a/components/header/podIndicator/styles.js
+++ b/components/header/podIndicator/styles.js
@@ -40,7 +40,7 @@ const styles = (theme) =>
     indicatorName: {
       textOverflow: "ellipsis",
       overflow: "hidden",
-      maxWidth: "200px",
+      maxWidth: "12rem",
       whiteSpace: "nowrap",
     },
     popover: {
--- a/components/header/podIndicator/styles.js
+++ b/components/header/podIndicator/styles.js
@@ -41,6 +41,7 @@ const styles = (theme) =>
       textOverflow: "ellipsis",
       overflow: "hidden",
       maxWidth: "200px",
+      whiteSpace: "nowrap",
     },
     popover: {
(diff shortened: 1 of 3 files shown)


components/header/podIndicator/styles.js
```diff
@@ -31,11 +31,17 @@ const styles = (theme) =>
...
+      display: "inline-flex",
       color: theme.palette.text.secondary,
       textTransform: "none",
       fontSize: "0.825rem",
       fontWeight: theme.typography.fontWeightRegular,
     },
+    indicatorName: {
+      textOverflow: "ellipsis",
+      overflow: "hidden",
+      maxWidth: "200px",
```
andydavison, line 43: I wasn't sure on the convention for units to use here — can switch to theme spacing/rem/etc.

ajacksified: I'd use rem; `10rem` should do it.

You'll also want a `white-space: nowrap` to prevent names with spaces creating a second line of text.

Screen Shot 2020-10-23 at 8 57 04 AM

Screen Shot 2020-10-23 at 8 58 02 AM

Lastly, this needs a bit of restructuring - because it's in the same `div` as the logo, long names push the icons out. The pod location indicator should be in a separate div, after everything else, so the content can flow above it:

Screen Shot 2020-10-23 at 9 00 07 AM

andydavison: yep, just noticed the wrapping issue (had a change in dev tools that wasn't committed 🤦‍♂️) Cool, will add that restructuring too 👍

andydavison: actually, if the name _isn't_ supposed to be in the same div as the logo, does it need truncating at all (unless it's more than full width of the screen?)

ajacksified: Only so the carat is obvious. Maybe truncate 50% of the screen?

ajacksified: (With a min-width, too.)

andydavison: The nav menu options are supposed to be full height, with icons vertically centred, they look odd shunted up to the top to accommodate a full width pod name(Files is focussed to show clickable area):

Screenshot 2020-10-23 at 14 39 04

Can obviously override or update prism styles, but not sure how much I should be fighting against it…

megoth: The PodIndicator is still part of the same div as logo, so it won't render the name underneath the icons.

I'm not 100% sure whether having the name go below the icons is the design we want though, maybe double-check with @KyraAssaad ?

andydavison: Got confirmation from @KyraAssaad on slack to go with truncation (rather than flowing underneath and disrupting alignment) _for now_, until she gets a chance to have a better think about how the nav layout works. Also confirmed the max-width chosen seems reasonable, and we're not too worried about that pushing the nav _slightly_

megoth (approved):

PR #148, opened 2020-10-27
Feat/hamburger menu

VirginiaBalseiro:
# Hamburger menu for small screens
In this PR:
- Slight refactor of the `header` component: `mainNav` is now its own component
- Added a hamburger menu only visible in small screens

### To test:
- Open Pod Browser, verify that the header still looks and work properly on desktop view
- Switch to mobile view, check that only the hamburger icon is visible in the header. 
- Click on the hamburger and verify that the links work properly

### Screenshots:
- Desktop:
image

- Mobile:
image
image

# Checklist


components/header/mainNav/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+import styles from "./styles";
+
+const useStyles = makeStyles((theme) => createStyles(styles(theme)));
+
+export default function MainNav() {
+  const classes = useStyles();
+
+  const bem = useBem(useStyles());
+  return (
+    <div className={classes.mainNav}>
```
megoth, line 36: for simplicity of reading, I would suggest only using one form of the result of useStyles, i.e. `classes.mainNav` --> `bem("mainNav")`.

megoth: Or if you want to have those two forms, at least reuse classes in useBem instead of invoking useStyles again.

components/header/menuDrawer/styles.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "menu-drawer-item__text": {
+      fontSize: theme.typography.body.fontSize,
+      fontWeight: theme.typography.fontWeightLight,
+    },
+    "menu-drawer-item__link": {
+      "text-decoration": "underline",
+    },
+    drawerContainer: {
+      borderRadius: "0px 0px 10px 10px",
```
megoth, line 39: Lets use the borderRadius from the theme on this one, `theme.shape.borderRadius`. It corresponds to 4 by default, so you can you do `0 0 ${theme.shape.borderRadius * 2.5} ${theme.shape.borderRadius * 2.5}` instead.

megoth: (But I think we might just want to use the default border radius - check with @KyraAssaad whether it should stay consistent with theme properties, or if we want it to be 10px.

components/header/menuDrawer/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+                  />
+                </ListItem>
+              </Link>
+            );
+          })}
+          <Divider />
+          <LogOutButton
+            className={bem("header-banner__user-menu-item-trigger")}
+          >
+            <span
```
megoth, line 101: Maybe have an icon for Log out like we do on the desktop version?

megoth: @KyraAssaad ^ ?

megoth (changes requested):
A couple of fixes, but nothing big.

VirginiaBalseiro:
@megoth I made all non-design related changes. For the icon in the logout option in the dropdown and the border radius, shall we have a separate discussion with @KyraAssaad and create the corresponding story if necessary?

megoth:
> @megoth I made all non-design related changes. For the icon in the logout option in the dropdown and the border radius, shall we have a separate discussion with @KyraAssaad and create the corresponding story if necessary?

I reckon improvements will be done through a more thorough review, so I think we can leave it as is for now.

megoth (approved):

kevin-inrupt (approved):

PR #152, opened 2020-10-28
Fix/table row keyboard events

VirginiaBalseiro:
This PR fixes: 
- Contact and Files tables are not keyboard accessible

Table rows in Contacts and Files can now be accessed via keyboard and the details drawer is opened when Enter key is pressed.


ajacksified (changes requested):
This opens the drawer - but focus should be moved to the context menu so that the actions can be tabbed through as well.

megoth (changes requested):
I'm not able to open the drawer for a contact using the enter key. It seems like the drawer closes right after it opens.

Maybe use `onKeyUp` instead?

VirginiaBalseiro:
> I'm not able to open the drawer for a contact using the enter key. It seems like the drawer closes right after it opens.
> 
> Maybe use `onKeyUp` instead?

Good catch! for whatever reason it was working for me locally but not on the Vercel preview 🤯 It should be fixed now!

megoth (approved):

ajacksified (approved):

kevin-inrupt (approved):

PR #157, opened 2020-10-28
Disallow modifying own permissions

ajacksified:
Screen Shot 2020-10-28 at 1 55 49 PM


components/resourceDetails/resourceSharing/agentAccess/index.jsx
```diff
@@ -132,48 +123,11 @@ export default function AgentAccess({
   const onSubmit = submitHandler(
     authenticatedWebId,
     webId,
-    setOpen,
```
megoth, line 135: I think this should still be here? `submitHandler` still takes that parameter, at least.

megoth (changes requested):

megoth (approved):

PR #161, opened 2020-10-29
Fix: add Profile link to hamburger menu in mobile

VirginiaBalseiro:
This PR adds the Profile link to the hamburger menu in mobile.

Matched the link style to the Logout link since we don't have "Profile" in the design, and made sense to put it after the divider along with Logout since it's part of the user menu items.

## Screenshot

image

### DIFF
--- a/components/header/menuDrawer/index.jsx
+++ b/components/header/menuDrawer/index.jsx
@@ -95,18 +95,36 @@ export default function MenuDrawer() {
             );
           })}
           <Divider />
-          <LogOutButton
-            className={bem("header-banner__user-menu-item-trigger")}
-          >
-            <span
+          <ListItem button key="profile" onClick={handleDrawerClose}>
+            <Link href="/profile" replace>
+              <span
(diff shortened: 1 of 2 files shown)


components/header/menuDrawer/index.jsx
```diff
@@ -95,18 +95,36 @@ export default function MenuDrawer() {
             );
           })}
           <Divider />
-          <LogOutButton
-            className={bem("header-banner__user-menu-item-trigger")}
-          >
-            <span
+          <ListItem button key="profile" onClick={handleDrawerClose}>
+            <Link href="/profile" replace>
+              <span
```
megoth, line 100: I think this should be an anchor-element (so we get semantically correct HTML).

megoth (approved):
Replace span with `a`-element if possible, otherwise it looks good.

megoth (approved):

kevin-inrupt (approved):

PR #164, opened 2020-10-30
Handling uploaded files that starts with a dash

megoth:
# Handling uploaded files that starts with a dash

With this PR we circumwent the problem with uploading files that starts with a dash by simply removing the initial dash.

### DIFF
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -21,7 +21,7 @@
 
 import React, { useContext, useState, useEffect } from "react";
 import T from "prop-types";
-import { getSourceUrl, overwriteFile } from "@inrupt/solid-client";
+import { overwriteFile } from "@inrupt/solid-client";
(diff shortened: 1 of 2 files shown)


components/addFileButton/index.jsx
```diff
@@ -43,8 +43,9 @@ export function handleSaveResource({
 }) {
   return async (uploadedFile) => {
     try {
-      const response = await overwriteFile(
-        joinPath(currentUri, encodeURIComponent(uploadedFile.name)),
+      const fileName = uploadedFile.name.replace(/^-/, "");
```
ajacksified, line 46: Doing this here would prevent the "Do you want to overwrite the existing file?" dialog from popping up, if `foo` exists and you try to upload `-foo`.

ajacksified (changes requested):

ajacksified (approved):

kevin-inrupt (approved):

PR #167, opened 2020-11-03
refactor to use people index everywhere

VirginiaBalseiro:
This PR fixes: usePeople should use People index to fetch profiles

- Rather than returning a list of profiles, `usePeople` now returns the people index dataset, which is then used to fetch the profiles where needed

### DIFF
--- a/src/addressBook/index.js
+++ b/src/addressBook/index.js
@@ -33,11 +33,10 @@ import {
   getUrl,
   getUrlAll,
   removeThing,
-  getSolidDataset,
   setThing,
 } from "@inrupt/solid-client";
 import { v4 as uuid } from "uuid";
-import { rdf, dc, acl, vcard, ldp, foaf, schema } from "rdf-namespaces";
+import { rdf, dc, acl, vcard, foaf, schema } from "rdf-namespaces";
 import {
   createResponder,
   defineDataset,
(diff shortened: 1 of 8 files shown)


megoth (approved):

src/addressBook/index.js
```diff
@@ -130,47 +129,16 @@ export async function getGroups(containerIri, fetch) {
...
-
-  if (contactsError) {
-    return error(contactsError);
-  }
-
-  const contactsIris = getUrlAll(
-    contactsResponse.dataset,
-    ldp.contains
-  ).map((url) => joinPath(url, INDEX_FILE));
+export async function getProfiles(peopleThings, fetch) {
```
ajacksified, line 132: Why rename this from `contacts` to `people`? It still looks generic to me.

src/addressBook/index.js
```diff
@@ -337,13 +305,8 @@ export function createContact(addressBookIri, contact, types) {
...
 }
 
-export async function findContactInAddressBook(addressBookIri, webId, fetch) {
-  const { response: people } = await getContacts(
-    foaf.Person,
-    addressBookIri,
-    fetch
-  );
-
+export async function findContactInAddressBook(peopleDataset, webId, fetch) {
```
ajacksified, line 308: Same here and everywhere else in this file - ideally this would be generic rather than use the word `people`, as it would also work for organizations, apps, etc.

ajacksified (changes requested):

ajacksified:
This PR goes a little beyond what we initially needed and reverts some recent changes. Per discussion with @VirginiaBalseiro, we'll close this and open a new PR which updates the logic for fetching contacts from reading the container to reading the index file.

PR #170, opened 2020-11-03
Use foaf.name as fallback to vcard.fn

ajacksified:
In user profiles, "Show contact", and the contacts list, this uses vcard.fn and falls back to foaf.name if unavailable. Eventually, this code should be replaced with the ability to read a fallback from the react sdk.

### DIFF
--- a/components/contactsList/index.jsx
+++ b/components/contactsList/index.jsx
@@ -83,7 +83,6 @@ function ContactsList() {
     addressBook
   );
   const profiles = useProfiles(people);
-  const formattedNamePredicate = vcard.fn;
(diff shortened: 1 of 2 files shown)


megoth:
I'm getting an error
bilde

megoth (approved):

kevin-inrupt:
Suggestion is to remove Name label and value from a Contact when displayed in read-only mode.

When a user's Profile is edited, the Name label and value needs to be displayed as that is where the Name is edited.

ajacksified:
Closed, recreating.

PR #172, opened 2020-11-04
tsConfig -> tsconfig (ts-jest deprecating the former)

megoth:
`ts-jest` is deprecating `tsConfig` in favor of `tsconfig`.


megoth:
Will fix this in the work I'm doing on improving test coverage

PR #178, opened 2020-11-06
Feat: bookmark links open parent container and resource details drawer

VirginiaBalseiro:
# Navigate to bookmarked resource

This PR fixes the issue of the "resource not supported" message when the clicking on a bookmarked resource that isn't a container, by navigating to the parent container, preselecting the row, and opening the drawer for the resource. 

# Checklist

### DIFF
--- a/components/containerTableRow/index.jsx
+++ b/components/containerTableRow/index.jsx
@@ -20,19 +20,20 @@
  */
 
 /* eslint-disable camelcase */
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { makeStyles, createStyles } from "@material-ui/styles";
 import { useBem } from "@solid/lit-prism-patterns";
 import { useRouter } from "next/router";
 import clsx from "clsx";
 import T from "prop-types";
+import { isContainer } from "@inrupt/solid-client";
 import { DETAILS_CONTEXT_ACTIONS } from "../../src/contexts/detailsMenuContext";
-import { isContainerIri } from "../../src/solidClientHelpers/utils";
 import PodLocationContext from "../../src/contexts/podLocationContext";
 import ResourceLink from "../resourceLink";
 import Bookmark from "../bookmark";
 import styles from "./styles";
 import { resourceContextRedirect } from "../../src/navigator";
+import { isContainerIri } from "../../src/solidClientHelpers/utils";
 
 export function ResourceIcon({ iri, bem }) {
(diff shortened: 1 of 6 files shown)


components/containerTableRow/index.jsx
```diff
@@ -64,13 +65,46 @@ export function handleAction(resourceIri, containerIri, router) {
...
+  container,
+  preselected,
+}) {
   const classes = useStyles();
   const bem = useBem(classes);
   const { name, iri } = resource;
   const router = useRouter();
   const { currentUri } = useContext(PodLocationContext);
-  const isActive = router.query.resourceIri === iri;
+  const isActive = router.query.resourceIri === iri || preselected;
```
ajacksified, line 85: Is a `preselected` parameter necessary? It seems as though you could just do a router redirect with the resource iri in the querystring.

megoth: Yes, I think this approach is better. Also, might be worth doing a "lookup" of the resource, i.e. do a `getResourceInfo` and check if the resource is a container or not from the response (by checking the URL that is returned - should verify whether it's a container or a non-container).

megoth (changes requested):

components/containerTableRow/index.jsx
VirginiaBalseiro: @ajacksified as per discussion I have removed `preselected` from the `isActive` condition.

@megoth I'm checking below whether it is a container using solid-client's `isContainer` in order to decide whether to redirect to the details drawer or not. Is this what you mean?

megoth: `isContainer` on the iri directly works in most situations, but I encountered an edge-case when I worked on this in the PodNavigator where users could type the URL to a folder _without_ the slash, and if we don't double-check by doing a lookup (i.e. fetch the resource using `useResourceInfo`) and use `isContainer` on the returned URL (which will always be "correct" in terms of having an ending slash or not).

VirginiaBalseiro: In this case we're getting the iri from the bookmark, which gets it from the container table row when the bookmark is created, which gets it from the dataset. Wouldn't it be overkill to fetch the resource info since it does not involve the user manually typing the iri?

megoth: Ah, yes, for bookmarks it probably doesn't make sense 👍

megoth (approved):

PR #179, opened 2020-11-09
Fix: add min and max width to input

VirginiaBalseiro:
This PR fixes: bookmark list search box exceeding screen width in small devices.

The input box now has a max width (like the Contacts search input box) and a width.


ajacksified (approved):
Ideally that `450` would come from a constant somewhere in prism - we should make sure it makes its way there.

kevin-inrupt (approved):

PR #180, opened 2020-11-10
Fix: move deleteFile logic out of access control

VirginiaBalseiro:
This PR fixes the delay when trying to delete a resource (getting an error when clicking immediately after details drawer is loaded).

- create new deleteResource function that deletes a file and its corresponding policy (if available)
- remove deleteFile from access control
- update/add tests

### DIFF
--- a/components/deleteResourceLink/index.jsx
+++ b/components/deleteResourceLink/index.jsx
@@ -19,18 +19,22 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext } from "react";
 import T from "prop-types";
 import { useDataset, useSession } from "@inrupt/solid-ui-react";
 import DeleteLink from "../deleteLink";
 import usePoliciesContainer from "../../src/hooks/usePoliciesContainer";
 import AlertContext from "../../src/contexts/alertContext";
 import { deleteResource } from "../../src/solidClientHelpers/resource";
-import { getPolicyUrl } from "../../src/accessControl/acp";
 
-export function createDeleteHandler(resourceIri, policyUrl, onDelete, fetch) {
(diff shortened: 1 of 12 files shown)


components/deleteResourceLink/index.jsx
```diff
@@ -39,18 +44,33 @@ export default function DeleteResourceLink({
...
 
-  const handleDelete = createDeleteHandler(accessControl, onDelete);
+  const { policiesContainer } = usePoliciesContainer();
+  const [policyUrl, setPolicyUrl] = useState(null);
 
-  if (resourceInfoError || accessControlError) {
-    return null;
-  }
+  useEffect(() => {
+    if (!policiesContainer || !resourceIri) return;
```
megoth, line 53: You only use policyUrl for the createDeleteHandler, so I think we can defer some of the logic in the `useEffect` here by moving the getPolicyUrl to `deleteResource` in resource.js.

components/deleteResourceLink/index.jsx
```diff
@@ -39,18 +44,33 @@ export default function DeleteResourceLink({
...
+    (async () => {
+      const { response: resource, error } = await getResource(
```
megoth, line 55: To optimize this a bit, we can use `useDataset` instead, which fetches a cached version of the resource, if available.

components/deleteResourceLink/index.jsx
```diff
@@ -39,18 +44,33 @@ export default function DeleteResourceLink({
...
+        resourceIri,
+        fetch
+      );
+      if (error) {
+        throw error;
```
megoth, line 60: I think we should render an error message instead of throwing an error here. Remember that thrown errors can result in a blank page unless their picked up properly.

megoth (changes requested):

megoth (approved):

ajacksified (approved):

PR #182, opened 2020-11-13
Fix: switch useDataset for useResourceInfo

VirginiaBalseiro:
This PR fixes bug: cannot view resources.

Fix: switch useDataset for useResourceInfo to fix error with non-turtle resources

### DIFF
--- a/src/solidClientHelpers/resource.js
+++ b/src/solidClientHelpers/resource.js
@@ -25,6 +25,7 @@ import {
   createThing,
   deleteFile,
   getSolidDataset,
+  getSourceUrl,
   getThing,
   saveSolidDatasetAt,
   setThing,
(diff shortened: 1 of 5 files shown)


src/solidClientHelpers/resource.js
```diff
@@ -107,12 +107,12 @@ export async function saveResource({ dataset, iri }, fetch) {
 }
 
 export async function deleteResource(resource, policiesContainer, fetch) {
-  const { dataset, iri } = resource;
+  const { info, iri } = resource;
```
megoth, line 110: Instead of packaging a custom object with `info` and `iri`, I think you should send the resourceInfo object, and use `getSourceUrl` to get the `iri`.

megoth (changes requested):

megoth (approved):

kevin-inrupt (approved):

PR #197, opened 2020-11-23
Added support for passing IdP using query param

megoth:
# Allows passing IdP using query param

Can pass Idp to the login provider using a query param, e.g. `?idp=https%3A%2F%2Finrupt.net%2F`.

# Checklist

### DIFF
--- a/components/login/provider/index.jsx
+++ b/components/login/provider/index.jsx
@@ -80,19 +80,19 @@ export function getErrorMessage(error) {
 
 export default function Provider({ defaultError }) {
   const buttonBem = Button.useBem();
-  const [providerIri, setProviderIri] = useState("https://inrupt.net");
   const { login } = useSession();
   const [loginError, setLoginError] = useState(defaultError);
   const theme = useTheme();
   const idp = useIdpFromQuery();
+  const [providerIri, setProviderIri] = useState(
+    Object.values(providers)[0]?.iri
+  );
   const loginFieldRef = createRef();
 
   useEffect(() => {
     if (idp) {
       setProviderIri(idp.iri);
-    }
-    if (idp && loginFieldRef.current) {
-      loginFieldRef.current.querySelector("input")?.focus();
+      loginFieldRef.current?.querySelector("input")?.focus();
     }
   }, [idp, loginFieldRef]);
 
--- a/components/login/provider/index.jsx
+++ b/components/login/provider/index.jsx
@@ -37,9 +37,8 @@ import { LoginButton, useSession } from "@inrupt/solid-ui-react";
 import { Button } from "@inrupt/prism-react-components";
 import { generateRedirectUrl } from "../../../src/windowHelpers";
 import getIdentityProviders from "../../../constants/provider";
-
 import { ERROR_REGEXES, hasError } from "../../../src/error";
-import useIdp from "../../../src/hooks/useIdp";
+import useIdpFromQuery from "../../../src/hooks/useIdpFromQuery";
 
 const providers = getIdentityProviders();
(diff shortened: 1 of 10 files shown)


ajacksified (commented):
Should this automatically submit, or do we want the user to click login? (Possibly a security question in the case of phishing.)

src/hooks/useIdp/index.js
```diff
@@ -0,0 +1,50 @@
+/**
```
ajacksified, line 1: I'm not sure this is so generic it belongs here instead of under the login page. I think it could also use a rename to make it clear - `useIdpFromQuery` - because otherwise it seems as though it might provide additional information.

src/hooks/useIdp/index.js
```diff
@@ -0,0 +1,50 @@
...
+import { useEffect, useState } from "react";
+import { useRouter } from "next/router";
+
+export default function useIdp() {
+  const [idp, setIdp] = useState(null);
+  const router = useRouter();
+
+  useEffect(() => {
+    const idpQuery =
+      typeof router.query.idp === "object"
```
ajacksified, line 31: `Array.isArray()` is clearer. Maybe it's worth having a `useQuery` hook so this code is reusable?

ajacksified (changes requested):

src/hooks/useIdp/index.js
megoth: I think `useQuery` might be to generic - what would it give us that `router.query.<whatever>` doesn't give us?

ajacksified: This code:

`typeof router.query.idp === "object" ? router.query.idp[0] : router.query.idp;`

would become:

`const idp = useQuery("idp");`

megoth: So something that guarantees that you only get one value from possible multiple query params values for the same key; Ok, I can create a hook for that.

components/login/provider/index.jsx
```diff
@@ -31,17 +33,30 @@ import { generateRedirectUrl } from "../../../src/windowHelpers";
...
   const { login } = useSession();
+  const idp = useIdpFromQuery();
+  const [providerIri, setProviderIri] = useState("https://inrupt.net");
+  const loginFieldRef = createRef();
+
+  useEffect(() => {
+    if (idp) {
+      setProviderIri(idp.iri);
+    }
+    if (idp && loginFieldRef.current) {
```
ajacksified, line 56: `loginFieldRef.current` could be moved below:

`loginFieldRef.current?.querySelector("input")?.focus();`

components/login/provider/index.jsx
```diff
@@ -31,17 +33,30 @@ import { generateRedirectUrl } from "../../../src/windowHelpers";
...
+export const TESTCAFE_ID_LOGIN_FIELD = "login-field";
 
 const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
 export default function Provider() {
   const bem = useBem(useStyles());
-  const [providerIri, setProviderIri] = useState("https://inrupt.net");
   const { login } = useSession();
+  const idp = useIdpFromQuery();
+  const [providerIri, setProviderIri] = useState("https://inrupt.net");
```
ajacksified, line 49: I know this is old code, but `"https://inrupt.net"` really should be in a constant (DEFAULT_PROVIDER), and even better if it used a provider from `getIdentityProviders`.

ajacksified (changes requested):

ajacksified (approved):

kevin-inrupt (approved):

PR #204, opened 2020-11-26
Feat/bookmark pod root

VirginiaBalseiro:
# Bookmark Pod root

This PR adds the functionality to bookmark a pod root, which is saved in bookmarks with the title {PodOwner}'s name.

This involved refactoring the Bookmark component to be used in the dropdown menu in the pod indicator. 
We can now pass a `menuItem` prop to the Bookmark component and it will return a List Item that can be used in this menu. 
It is also possible to pass `withText` and it will return a button we can use anywhere, but with text next to the icon. The strings can also be passed to the component. This is in case we need to use the bookmark button but with text anywhere, and not necessarily use the Material UI List Item styles (for example in the Details drawer).

Tests were also updated.

# Checklist


components/bookmark/index.test.jsx
```diff
@@ -41,7 +42,65 @@ describe("Bookmark", () => {
...
     expect(asFragment()).toMatchSnapshot();
   });
+
+  test("it renders a bookmark icon and text if withText prop is true", async () => {
+    const { asFragment } = renderWithTheme(
+      <BookmarkContext.Provider value={(bookmarks, setBookmarks)}>
+        <Bookmark iri={iri} withText />
+      </BookmarkContext.Provider>
+    );
+    const text = screen.getByTestId("bookmark-text");
```
megoth, line 52: I think you can might as well get `getByTestId` from the resulting object from `renderWithTheme` (e.g. `const { asFragment, getByTestId }`. Don't think there's much difference, but then you have to import one less dependency.

megoth (approved):
Would've liked specific tests for `bookmarkButton` and `bookmarkText`, but see that they're mostly covered by other tests. If you could get that Function coverage up from 50 to 100 on `bookmarkButton`, and Branch coverage from 83.33 to above 90 for `bookmarkText`, that would be great.

I'll approve this PR thought, as it's meeting the requirement of test coverage.

VirginiaBalseiro:
> Would've liked specific tests for `bookmarkButton` and `bookmarkText`, but see that they're mostly covered by other tests. If you could get that Function coverage up from 50 to 100 on `bookmarkButton`, and Branch coverage from 83.33 to above 90 for `bookmarkText`, that would be great.
> 
> I'll approve this PR thought, as it's meeting the requirement of test coverage.

Added separate test files with additional cases. The 50% coverage in the BookmarkButton component was complaining about a default prop 🤦‍♀️ I added a test for that to keep it happy.

megoth (approved):

megoth (approved):
The disable-while-updating for the bookmark button kind of work, but is barely noticeable when it's part of the Navigator table. I suspect we might want to improve that even more, but I'll leave that decision for Kyra.

VirginiaBalseiro:
> The disable-while-updating for the bookmark button kind of work, but is barely noticeable when it's part of the Navigator table. I suspect we might want to improve that even more, but I'll leave that decision for Kyra.

I only changed it for the dropdown menu item in this PR, so we can have a separate conversation about the bookmark icon in the navigator table for sure. The main goal for disabling this button is to prevent the user clicking multiple times and triggering more add/remove actions.

kevin-inrupt (approved):

PR #205, opened 2020-11-27
Fix: disallow existing or own webId in permissions

VirginiaBalseiro:
This PR fixes bug: users can still overwrite their own permissions by re-adding their webId.

We now display an error when the users tries to add themselves to the permissions, or a webId that is also in the permissions for that resource.

### DIFF
--- a/components/addPermissionUsingWebIdButton/index.jsx
+++ b/components/addPermissionUsingWebIdButton/index.jsx
@@ -108,7 +108,6 @@ export default function AddPermissionUsingWebIdButton({
 
   useEffect(() => {
     if (!accessControl) {
-      setPermissions(null);
       return;
     }
     accessControl.getPermissions().then((normalizedPermissions) => {
--- a/components/addPermissionUsingWebIdButton/index.jsx
+++ b/components/addPermissionUsingWebIdButton/index.jsx
@@ -21,7 +21,7 @@
 
 /* eslint-disable react/jsx-props-no-spreading */
 
-import React, { useContext, useState } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import T from "prop-types";
 import { createStyles, makeStyles, Popover } from "@material-ui/core";
 import { InputGroup, Label, Message } from "@inrupt/prism-react-components";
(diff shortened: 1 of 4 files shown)


components/agentSearchForm/index.jsx
```diff
@@ -57,6 +64,13 @@ function AgentSearchForm({
...
         <Message variant="invalid">Please provide a valid WebID</Message>
       ) : null}
+      {existingWebId ? (
+        <Message variant="error">
+          {session.info.webId === existingWebId
+            ? "You cannot overwrite your own permissions."
+            : `The WebID ${existingWebId} is already in your permissions.`}
+        </Message>
+      ) : null}
       <SimpleInput
```
megoth, line 74: You also want the input field to show being invalid - use the modifier `invalid` to do that (or with the React component, use the property `variant` with `invalid`).

megoth (changes requested):

megoth (approved):

megoth (approved):

megoth:
The only change you might want to look at is cleaning up the validation error messages (like if you try to add your own WebId, then another WebId that already exist, both error messages are shown). But I think this might be solved better (and more consistently) by making sure we get a better, generic validation system in place.

VirginiaBalseiro:
> The only change you might want to look at is cleaning up the validation error messages (like if you try to add your own WebId, then another WebId that already exist, both error messages are shown). But I think this might be solved better (and more consistently) by making sure we get a better, generic validation system in place.

Good catch, thanks! I added this now, if you want to check again.

components/addPermissionUsingWebIdButton/index.jsx
```diff
@@ -99,12 +99,23 @@ export default function AddPermissionUsingWebIdButton({
...
+  const [permissions, setPermissions] = useState(null);
   const classes = useStyles();
   const [disabled, setDisabled] = useState(false);
   const { accessControl } = useContext(AccessControlContext);
   const [agentId, setAgentId] = useState("");
   const [dirtyForm, setDirtyForm] = useState(false);
 
+  useEffect(() => {
+    if (!accessControl) {
+      setPermissions(null);
```
ajacksified, line 111: With `permissions` defaulting to `null` anyways, is this line necessary?

components/agentSearchForm/index.jsx
```diff
@@ -57,6 +72,16 @@ function AgentSearchForm({
       {invalidWebIdField ? (
         <Message variant="invalid">Please provide a valid WebID</Message>
       ) : null}
+      {isPodOwner ? (
```
ajacksified, line 75: `isPodOwner && (<Message />)` is a slightly cleaner way to handle conditional rendering than a ternary.

components/addPermissionUsingWebIdButton/index.jsx
VirginiaBalseiro: Looks like it is not 👍

kevin-inrupt (approved):

components/addPermissionUsingWebIdButton/index.jsx
megoth: I think it's good to have in cases where permissions gonfrom null->something and then something->null again

VirginiaBalseiro: In this case though, it's when accessControl is not available, so it won't change to non-null, right? I might be missing something.

megoth: It might be going from something to null if you open a resource that you have Control access to first, and then to one that you don't have Control access.

But I'm not 100% on the flow there, so I might be wrong.

VirginiaBalseiro: When logging out the permissions, they don't seem to change when I add/remove this line going from a resource with control access to one without and viceversa.

ajacksified (approved):

PR #207, opened 2020-11-30
Added default content-type to uploaded files

megoth:
# Added default content-type to uploaded files

Sets `application/octet-stream` as fallback for files that the browser are not able to figure out content type for.

Also improved test coverage.

### DIFF
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -48,11 +48,14 @@ export function handleSaveResource({
   return async (uploadedFile) => {
     try {
       const fileName = normalizeSafeFileName(uploadedFile.name);
+      const { type } = uploadedFile;
+      Object.defineProperty(uploadedFile, "type", {
+        get: () => type || "application/octet-stream",
+      });
       await overwriteFile(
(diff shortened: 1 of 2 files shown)


ajacksified:
> Sets application/octet-stream as fallback for files that the browser are not able to figure out content type for.

It seems as though there were some reservations about this; and if we _are_ going to use this as a default, this should be in the SDK imo.

megoth:
> > Sets application/octet-stream as fallback for files that the browser are not able to figure out content type for.
> 
> It seems as though there were some reservations about this; and if we _are_ going to use this as a default, this should be in the SDK imo.

I agree with the sentiment, but wonder whether we should have this as an interim solution until we have a proper solution in the SDK?

Vinnl:
@megoth https://github.com/inrupt/solid-client-js/blob/4a1556fd333f3ce59fb345aa83c30082f2fd97a6/CHANGELOG.md#bugs-fixed :)

ajacksified:
Closing - will be handled another way with a solid-client upgrade in another ticket.

PR #223, opened 2020-12-15
Fix: use vcard.url instead of foaf.openid for new contacts

VirginiaBalseiro:
- with new contacts, create a 'blank' node under `vcard.url`
- read from `foaf.openid` and `vcard.url`
- update tests

### DIFF
--- a/__testUtils/mockPersonResource.js
+++ b/__testUtils/mockPersonResource.js
@@ -19,7 +19,12 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { addStringNoLocale, addUrl, mockThingFrom } from "@inrupt/solid-client";
+import {
+  addStringNoLocale,
+  addUrl,
+  asUrl,
+  mockThingFrom,
+} from "@inrupt/solid-client";
 import { vcard, foaf, rdf } from "rdf-namespaces";
 import { chain } from "../src/solidClientHelpers/utils";
 import { packageProfile } from "../src/solidClientHelpers/profile";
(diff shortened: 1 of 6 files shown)


components/contactsList/index.jsx
```diff
@@ -124,10 +124,12 @@ function ContactsList() {
...
-  const contacts = profiles.map((p) => ({
-    thing: p,
-    dataset: addressBook,
-  }));
+  const contacts = profiles.map((p) => {
+    return {
+      thing: p,
+      dataset: addressBook,
+    };
+  });
```
ajacksified, line 132: Could be cleaned up with an implicit return: 

```
const contacts = profiles.map((thing) => ({ thing, dataset: addressBook });
```

ajacksified (commented):
Looks good, just a minor comment.

ajacksified (approved):

__testUtils/mockPersonResource.js
```diff
@@ -29,14 +34,28 @@ export const aliceName = "Alice";
 export const aliceNick = "A";
 export const alicePhoto = "http://example.com/alice.jpg";
 
+const VCARD_WEBID_PREDICATE = "https://www.w3.org/2006/vcard/ns#WebId";
+
+export function mockWebIdNode(webId) {
+  const dataset = chain(
```
megoth, line 40: I would not call this `dataset`, since that can be confused with the type `SolidDataset`. Maybe `webIdNode` or `webIdThinng` would be better?

src/addressBook/index.js
```diff
@@ -337,13 +369,13 @@ export function createContact(addressBookIri, contact, types) {
...
   );
-
-  const dataset = [...emails, ...addresses, ...telephones].reduce(
+  let dataset = [...emails, ...addresses, ...telephones].reduce(
     (acc, { thing }) => {
       return setThing(acc, thing);
     },
     person
   );
+  dataset = webIdNode && setThing(dataset, webIdNode);
```
megoth, line 378: I'm a little concerned about this line, as it makes the overall logic a bit hard to follow for me. I reckon if `webIdNode` is falsy, it will simply return `false` for dataset? I cannot see any test that back up this scenario, or am I misunderstanding this?

megoth (changes requested):
Just want to clarify a couple of things, then we should be good.

src/addressBook/index.js
VirginiaBalseiro: I can't remember exactly why I added that check tbh. Removed it since the webIdNode is never undefined (there's always a webId when we add a contact, and if it's invalid it will fail to fetch the profile and won't add the contact anyway).

kevin-inrupt (approved):

megoth (approved):

megoth:
Btw, I forgot this earlier, but the ShEx shapes we must also be updated to reflect this change.

megoth (approved):

PR #230, opened 2021-01-04
fix: focus on upload button and trigger file browser opening on enter

VirginiaBalseiro:
This PR adds:
- The "Upload File" button is now focusable
- The file browser opening is triggered by pressing "Enter" 

### DIFF
--- a/components/addFileButton/index.jsx
+++ b/components/addFileButton/index.jsx
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect, useRef, useState } from "react";
 import T from "prop-types";
 import { overwriteFile } from "@inrupt/solid-client";
 import { useSession } from "@inrupt/solid-ui-react";
@@ -30,6 +30,8 @@ import { joinPath } from "../../src/stringHelpers";
 
 const TESTCAFE_ID_UPLOAD_BUTTON = "upload-file-button";
 const TESTCAFE_ID_UPLOAD_INPUT = "upload-file-input";
+const TESTCAFE_ID_UPLOAD_LABEL = "upload-file-label";
+
 export const DUPLICATE_DIALOG_ID = "upload-duplicate-file";
 
 function normalizeSafeFileName(fileName) {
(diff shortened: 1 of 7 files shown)


components/addFileButton/index.jsx
```diff
@@ -216,23 +219,36 @@ export default function AddFileButton({ className, onSave, resourceList }) {
   }, [confirmationSetup, confirmed, onConfirmation, file, open]);
 
   return (
-    // eslint-disable-next-line jsx-a11y/label-has-associated-control
-    <label
-      data-testid={TESTCAFE_ID_UPLOAD_BUTTON}
-      disabled={isUploading}
+    <button
```
megoth, line 222: The combination of button > label > input is bad. Wouldn't adding `tabIndex={0}` to the label be a better option, and add a onKeyUp event handler that triggers the input field when Enter is clicked? E.g.

```
const inputRef = createRef();
```

In combination with:

```
    <label
      data-testid={TESTCAFE_ID_UPLOAD_BUTTON}
      disabled={isUploading}
      className={className}
      tabIndex={0}
      onKeyUp={(event) => {
        if (event.code !== "Enter") return;
        inputRef.current.click();
      }}
      role="button"
    >
      {isUploading ? "Uploading..." : "Upload File"}
      <input
        data-testid={TESTCAFE_ID_UPLOAD_INPUT}
        type="file"
        style={{ display: "none" }}
        onClick={(e) => {
          e.target.value = null;
        }}
        onChange={onFileSelect}
        ref={inputRef}
      />
    </label>
```

(I've tried this out, and ESLint returns an error "jsx-a11y/no-noninteractive-element-to-interactive-role" for `role="button"`; I think it's better to just ignore that one for this case.)

This should also allow you to test out the event handler doing `userEvent.type(getByTestId(TESTCAFE_ID_UPLOAD_BUTTON), "{enter}")`, perhaps with an event listener added to the click event of the input field itself to see that it has been triggered.

megoth (changes requested):

components/addFileButton/index.jsx
VirginiaBalseiro: @megoth changed it back to label and ignored two warnings for interactive elements (tabIndex and interactions).

megoth (approved):

kevin-inrupt (approved):

PR #231, opened 2021-01-06
fix: handle deletion of current folder

VirginiaBalseiro:
This PR fixes bug: pod indicator displays "unknown" after deleting current folder

The cause was redirecting to the deleted folder: `stripQueryParams` did not work in the case of the current folder because there are no query params to remove and we need to go up one level. When trying to fetch the pod owner for the resource requested, it cause an error which set the pod owner web to null.

### DIFF
--- a/components/containerDetails/index.test.jsx
+++ b/components/containerDetails/index.test.jsx
@@ -87,7 +87,7 @@ describe("ContainerDetails", () => {
     beforeEach(() => onDeleteCurrentContainerFn());
 
     test("it calls mutate", () => expect(mutate).toHaveBeenCalled());
-    test("it calls handleCloseDrawerFn", () =>
+    test("it calls handleRedirectToParentContainerFn", () =>
       expect(handleRedirectToParentContainerFn).toHaveBeenCalled());
   });
 });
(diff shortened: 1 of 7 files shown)


components/containerDetails/index.test.jsx
```diff
@@ -70,4 +82,12 @@ describe("ContainerDetails", () => {
     test("it calls handleCloseDrawerFn", () =>
       expect(handleCloseDrawerFn).toHaveBeenCalled());
   });
+
+  describe("when onDeleteCurrentContainer is called", () => {
+    beforeEach(() => onDeleteCurrentContainerFn());
+
+    test("it calls mutate", () => expect(mutate).toHaveBeenCalled());
+    test("it calls handleCloseDrawerFn", () =>
+      expect(handleRedirectToParentContainerFn).toHaveBeenCalled());
```
megoth, line 91: It would be even better if you checked what the spy has been called with.

megoth (approved):
One minor suggestion, otherwise looks good

megoth (approved):

kevin-inrupt (approved):

PR #239, opened 2021-01-25
Introducing Group API

megoth:
# Introducing Group API

This PR adds CRUD methods for groups, and tries to standardize some of the concepts into models.

The idea is when a developer sees `group`, `addressBook`, or `contact` in the code, they should be able to assume that it refers to a specific model.

I've repurposed some existing methods from `src/addressBook` into `src/models/addressBook` and `src/models/contact` so that they can live on as model methods. The remaining code in `src/addressBook` should either be moved to the appropriate model (I didn't move everything, as I focused on what I needed for groups for this PR), replaced with better methods, or removed altogether.

I have not added support for `description` in groups, as I think this could be handled with Solid UI components.

# Checklist


__testUtils/mockGroup.js
```diff
@@ -0,0 +1,104 @@
...
+  groupThingUrl
+) {
+  return chain(
+    dataset,
+    (d) => setThing(d, mockGroupThing(name, groupThingUrl)),
+    (d) => setThing(d, mockIndexThing(addressBook, groupThingUrl))
+  );
+}
+
+// export function addGroupsToAddressBook(addressBook, groupsToAdd) {
```
ajacksified, line 62: Should these commented lines be deleted?

src/models/addressBook/index.js
```diff
@@ -0,0 +1,180 @@
...
+  saveSolidDatasetAt,
+  setThing,
+} from "@inrupt/solid-client";
+import { acl, dc, foaf, rdf, schema, vcard } from "rdf-namespaces";
+import { joinPath } from "../../stringHelpers";
+import { chain } from "../../solidClientHelpers/utils";
+import { ERROR_CODES, isHTTPError } from "../../error";
+
+/**
+ * @typedef AddressBook
```
ajacksified, line 40: I'd prefer not to sprinkle typedoc through the codebase inconsistently.

src/models/addressBook/index.js
```diff
@@ -0,0 +1,180 @@
...
+export const NAME_EMAIL_INDEX_PREDICATE = vcardExtras("nameEmailIndex");
+export const NAME_GROUP_INDEX_PREDICATE = vcardExtras("groupIndex");
+export const VCARD_WEBID_PREDICATE = vcardExtras("WebId");
+export const INDEX_FILE = "index.ttl";
+export const PEOPLE_INDEX_FILE = "people.ttl";
+export const GROUPS_INDEX_FILE = "groups.ttl";
+export const PERSON_CONTAINER = "Person";
+export const GROUP_CONTAINER = "Group";
+
+const person = {
```
ajacksified, line 63: This model is still referring to "person" and "group". Per our conversations, these concepts should not exist in this file and should be lazily created when a person or group is added to an address book for the first time.

ajacksified (changes requested):

src/models/addressBook/index.js
megoth: I wanted to use it in models as a reference, to lessen the chance of inconsistent types.

megoth: I can of course not use JSDoc syntax if you prefer? But I still want some documentation in the "head" of each model module.

src/models/addressBook/index.js
megoth: Ok, think I've been able to move out all "person" and "group" related info from AddressBook now - everything is now handled by the Contact model, including a method that updates the AddressBook main index when a group is created and group index doesn't exist from before.

Do note that this violates the constraints described in https://github.com/inrupt/pod-browser/blob/master/shapes/contacts/address-book.shex, which states that `vcard:nameEmailIndex` and `vcard:groupIndex` should be set. That said, I do like the lazy-creation approach, so if anything we can discuss with the SolidOS/Data Browser team whether we should allow that flexibility.

__testUtils/mockAddressBook.js
```diff
@@ -0,0 +1,81 @@
...
+  const containerIri = options.containerIri || defaultContainerIri;
+  const owner = options.owner || aliceWebIdUrl;
+  const indexUrl =
+    options.indexUrl || getAddressBookIndexDefaultUrl(containerIri);
+  const groupsUrl =
+    options.groupsUrl ?? getContactIndexDefaultUrl(containerIri, vcard.Group);
+  const peopleUrl =
+    options.peopleUrl ?? getContactIndexDefaultUrl(containerIri, foaf.Person);
+  const mainIndexIri = `${indexUrl}#this`;
+  debugger;
```
ajacksified, line 48: Remove debugger - surprised the linter allows this. Also, groups and people should be able to be removed from the mock.

src/models/addressBook/index.js
```diff
@@ -0,0 +1,132 @@
...
+      mainIndexUrl,
+      newAddressBook.dataset,
+      {
+        fetch,
+      }
+    );
+    const thing = getThing(dataset, `${getSourceUrl(dataset)}#this`);
+    return { containerIri, dataset, thing };
+  } catch (error) {
+    if (isHTTPError(error, ERROR_CODES.UNAUTHORIZED)) {
```
ajacksified, line 127: Since this try/catch only cares whether it's HTTP or it re-throws, you should be able to only wrap the `saveSolidDatasetAt` call.

src/models/contact/index.js
```diff
@@ -0,0 +1,129 @@
...
+ * @type {object}
+ * @property {object} dataset - The dataset that the thing lives in
+ * @property {object} thing - The contact itself, be that a person, a group, or something else
+ */
+
+/* Model constants */
+export const NAME_EMAIL_INDEX_PREDICATE = vcardExtras("nameEmailIndex");
+export const NAME_GROUP_INDEX_PREDICATE = vcardExtras("groupIndex");
+export const INDEX_FILE = "index.ttl";
+export const PEOPLE_INDEX_FILE = "people.ttl";
```
ajacksified, line 51: GROUP constants should live in models/group, and PEOPLE constants should live in a models/personContact (which doesn't exist.) (While you're at it, rename models/group to models/groupContact for clarity.)

I would suggest removing the type_map and indexes, and create `saveGroupInAddressBook()` in models/group and `savePersonInAddressBook()` methods in a new `models/personContact`, which call `addContactIndexToAddressBook` with the appropriate type. That way, `contact` doesn't know or care about types, the actual typed contacts do.

ajacksified (changes requested):

__testUtils/mockAddressBook.js
megoth: I'll see what I can do about the mocking methods - they're a bit convoluted right now, so yeah, could probably clear up a bit there.

ajacksified (approved):

PR #243, opened 2021-01-29
Allowing visitors to change their own permissions

megoth:
This solutions is not complete though, as the details won't reload properly if you remove your own Control access for example.

But since this component is about to be replaced, I advice against investing to much time into this component.


components/permissionsForm/index.test.jsx
```diff
@@ -21,15 +21,31 @@
...
 describe("PermissionsForm", () => {
+  const authProfile = mockProfileAlice();
+  const visitorIri = "http://some-random-pod.com/";
+
+  beforeEach(() => {
+    jest
+      .spyOn(routerFns, "useRouter")
+      .mockReturnValue({ query: { iri: visitorIri } });
+    mockedAuthenticationProfile.mockReturnValue({ data: authProfile });
+  });
```
ajacksified, line 47: Tests should be updated to reflect the updated `disabled` logic - such as a negative case.

ajacksified (changes requested):

components/permissionsForm/index.test.jsx
megoth: This is handled in https://github.com/inrupt/pod-browser/pull/243/files#diff-fb36631f81b590b8107939e2bc4400e75b8dff0f751d0d0a67f26973f157f40dR180-R194. I did consider whether to have a test directly in permissionsForm.test.jsx as well, but didn't think it was worth doing, since it's covered, and it's a component we'll phase out soon anyway.

But ok, I can write an extra test for it.

megoth: Tests written

ajacksified (approved):

kevin-inrupt (approved):

src/accessControl/acp/index.js
```diff
@@ -201,18 +201,35 @@ function ensureAccessControl(policyUrl, datasetWithAcr, changed) {
 }
 
 function ensureApplyControl(policyUrl, datasetWithAcr, changed) {
-  const accessControls = acp.getAllControl(datasetWithAcr);
+  let accessControls = [];
+  try {
+    accessControls = acp.getAllControl(datasetWithAcr);
+  } catch (error) {
+    // TODO: Handle this error
+    // It doesn't seem to affect the outcome though, so we'll leave it like this for now
```
ajacksified, line 209: Yikes. What are the possible errors here?

src/accessControl/acp/index.js
```diff
@@ -201,18 +201,35 @@ function ensureAccessControl(policyUrl, datasetWithAcr, changed) {
...
+  }
+  let acr = datasetWithAcr;
+  try {
+    acr = acp.setControl(
+      datasetWithAcr,
+      chain(acp.createControl(), (ac) => acp.addPolicyUrl(ac, policyUrl))
+    );
+  } catch (error) {
+    // TODO: Handle this error
+    // It doesn't seem to affect the outcome though, so we'll leave it like this for now
```
ajacksified, line 228: Same here - how does `setControl` failing not affect the outcome? What are the possible errors?

src/accessControl/acp/index.js
megoth: Good question - It seems that the `acp.getAllControl` tries to to a `getThingAll` on the dataset, but for some reason it fails. I couldn't find any good solution, and I think we probably want to rework this anyway when we get the high level API for the Solid client, so thought it wasn't worth it to debug deeper.

src/accessControl/acp/index.js
megoth: Same answer as above.

I could debug further to figure out exactly what's going on, but I'm not sure it's worth the time it will take.

src/accessControl/acp/index.js
ajacksified: Can you update the comment with this explanation?

megoth:
I'll create a bug ticket for the remaining, unresolved bugs, since it relies on a bug fix in Solid client. The original bug for this PR is fixed though, so I'll merge it now.

PR #246, opened 2021-02-03
Update ShEx shapes to be more self-documented

pmcb55:
Just some comments for discussion - it would be nice to be able to share this file (with explanatory Turtle perhaps) with external parties who may be looking at modelling these ideas using Upper/Mid-Level Ontologies, but it would need some clarifications first I think... (and comments explaining 'missing' terms, etc.)


megoth:
I've added some comments that should make each ShEx shape more capable of documenting how they're used

shapes/bookmarks/bookmark.shex
```diff
@@ -3,6 +3,9 @@ PREFIX foaf:      <http://xmlns.com/foaf/0.1/>
 PREFIX terms:     <http://purl.org/dc/terms/>
 PREFIX xsd:       <https://www.w3.org/2001/XMLSchema#>
 
+# This shape describes some the information we store a bookmark. The
+# resource it is used describes multiple bookmarks.
+
```
pmcb55, line 8: I don't know what the 2nd sentence is trying to say - is it trying to say a single Resource can have multiple bookmark instances referring to it (e.g., multiple bookmark instances with different `dcterms:created` and `dcterms:title` values, but all with the same `bookmark:recall` value)...?
```suggestion
# This shape describes some of the information we store for a bookmark. The
# resource it is used describes multiple bookmarks.

```

shapes/contacts/address-book.shex
```diff
@@ -3,9 +3,22 @@ PREFIX dc:    <http://purl.org/dc/elements/1.1/>
 PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
 PREFIX xsd:   <http://www.w3.org/2001/XMLSchema#>
 
+# This shape describes some of the information we store for an address book. The
+# resource it is used for is to describe a single address book. We use the id
```
pmcb55, line 7: I think this wording is a little clearer (and shorter):
```suggestion
# This shape describes some of the information we store for an address book. It's
# intended to represent a single address book. We use the ID
```

shapes/bookmarks/bookmark.shex
megoth: The intention is to say that the resource where we use this shape for contains multiple bookmarks.

shapes/contacts/address-book.shex
```diff
@@ -3,9 +3,22 @@ PREFIX dc:    <http://purl.org/dc/elements/1.1/>
...
+# "#this" to refer to the subject of the address book.
```
pmcb55, line 8: This could use a bit more explanation I think (simple Turtle snippets are always your friend there I think :) !). I assume `#this` is a fragment identifier, and do you mean that you always apply (just as a **_naming convention_**) `#this` to the end of the RDF Subject IRI for AddressBook instances?
But why? Wouldn't that mean I can't dereference that AddressBook instance directly from my browser, i.e., if the instance was `https://pod.com/userX/public/myAddressBook#this`, then my browser will only request `https://pod.com/userX/public/myAddressBook`.
So I think I'm misinterpreting what is meant here by "the subject of the address book".

shapes/contacts/group-index.shex
```diff
@@ -1,7 +1,20 @@
 PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
-PREFIX xsd: <https://www.w3.org/2001/XMLSchema#>
+PREFIX xsd:   <https://www.w3.org/2001/XMLSchema#>
+
+# This shape describes part of the information we store when adding a group to
+# an address book. The resource it is used for is an index of groups. There will
```
pmcb55, line 5: I would say "The resource it is used for is an index of groups", since it could be reused in a million different ways by other people. You could just add that as an _example_ of where it _could_ be used, but if so I'd suggest rewording it slightly to something like "could be used to represent a group within an index of groups" (but then what's the difference between "an index of groups" and "a collection of groups", or "a container containing groups"...?).

shapes/contacts/group-index.shex
```diff
@@ -1,7 +1,20 @@
...
+# be multiple groups added in this resource, and each referring to the group
+# resource where there are more information on the group (e.g. members).
+
+# This model is originally developed as part of the
```
pmcb55, line 9:
```suggestion
# This model was originally developed as part of the
```

shapes/contacts/address-book.shex
```diff
@@ -3,9 +3,22 @@ PREFIX dc:    <http://purl.org/dc/elements/1.1/>
...
+
+# This model is originally developed as part of the
```
pmcb55, line 10:
```suggestion
# This model was originally developed as part of the
```

shapes/contacts/address-book.shex
megoth: Yeah, you will have to excuse my miswording in this, english is not my native language and I'm a bit out of training wrt using the right, formal words to describe these concepts.

The use of `#this` as a naming convention comes from SolidOS/Data Browser, where `#this` is given special treatment if found in a resource (i.e. it will render the contacts-pane when finding #this and seeing it's of type `vcard:AddressBook`).

shapes/contacts/group.shex
```diff
@@ -1,9 +1,22 @@
 PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
 PREFIX xsd:   <http://www.w3.org/2001/XMLSchema#>
 
+# This shape describes some of the information we store when adding a group to
+# an address book. The resource it is used for is to describe the group itself
+# (meaning the resource should only contain one group). We use the id "#this"
+# to refer to the subject of the group.
+
+# This model is originally developed as part of the
```
pmcb55, line 9:
```suggestion
# This model was originally developed as part of the
```

shapes/contacts/group.shex
```diff
@@ -1,9 +1,22 @@
 PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
 PREFIX xsd:   <http://www.w3.org/2001/XMLSchema#>
 
+# This shape describes some of the information we store when adding a group to
+# an address book. The resource it is used for is to describe the group itself
```
pmcb55, line 5: Again, overly specific about where this Shape would be applicable...

shapes/contacts/address-book.shex
pmcb55: Ok, no problem at all. The example Turtle files you've add are perfect for demonstrating the usage and intent of `#this`. So I'd just reference the local directory containing those files, something like this:
```suggestion
# "#this" to refer to the subject of the address book (see example Turtle files in ????).
```

shapes/bookmarks/bookmark.shex
pmcb55: Ah Ok - that's certainly clearer. So I'd suggest changing the 2nd sentence here to be something like this:
"Multiple instances conforming to this shape are intended to be contained together within a single resource - in other words, a single Solid resource could contain a collection of these bookmark instances, as opposed to having a dedicated Solid resource per bookmark instance."

shapes/contacts/people-index.shex
```diff
@@ -1,7 +1,22 @@
 PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
 PREFIX xsd:   <http://www.w3.org/2001/XMLSchema#>
 
+# This shape describes part of the information we store when adding a person to
+# an address book. The resource it is used for is an index of people. There will
+# be multiple people added in this resource, and each referring to the person
+# resource where there are more information on the person (e.g. photo, email,
+# etc).
```
pmcb55, line 8:
```suggestion
# be multiple people added to this resource, with each referring to the person
# resource where there may be more information on that person (e.g., photo, email,
# etc).
```

shapes/README.md
```diff
@@ -1,106 +1,153 @@
 # Shapes
 
+## Indexes
+
+We use the term 'index' throughout our modelling to refer to a single RDF
+resource that acts as a container for multiple 'contained' instances, where each
+contained instance may refer to an external resource, and may also contain
+copies of (some of the) data from that external resource 
+****
+(e.g., an instance may copy fields from an external resource, plus have a link
```
megoth, line 10: Perhaps remove this comment?

shapes/contacts/example/Group/Group_1.ttl
```diff
@@ -0,0 +1,13 @@
+@prefix :       <#>.
+@prefix vcard:  <http://www.w3.org/2006/vcard/ns#>.
+
+:this
+    a vcard:Group;
+    vcard:fn "Group 1";
+    vcard:hasMember <../Person/67fca8b3-c74a-499e-b00f-67c543fa533f/index.ttl#this>.
```
megoth, line 7: Didn't you have a change that points to another example file that exists?

shapes/contacts/example2/group/group3-soccer-team.ttl
```diff
@@ -0,0 +1,20 @@
+prefix :       <#>
+prefix vcard:  <http://www.w3.org/2006/vcard/ns#>
+
+:this
+    a vcard:Group ;
+    vcard:fn "Group 1" ;
+    vcard:hasMember <../Person/2d73ccc1-9d76-4398-8840-3c23a8ddf04c/index.ttl#this> ;
```
megoth, line 7: Maybe better to use `vcard:hasMember <person1> , <person2> .` instead of `vcard:hasMember <person1> ; vcard:hasMember <person2> .`?

megoth (approved):
I think this is a big step in improving the documentation, so I'm happy to approve as-is. But left a couple of suggestions for improvements, but not strictly necessary.

shapes/README.md
pmcb55: Oh yeah, that was an oversight!

shapes/contacts/example/Group/Group_1.ttl
pmcb55: Yeah, I've replaced `/example` now with what was in `/example2`, so this problem should be gone.

shapes/contact/group.shex
```diff
@@ -0,0 +1,26 @@
...
+# This model was originally developed as part of the
+# SolidOS/Data Browser project, and we strive to keep it interoperable with
+# their work. Some custom terms have been introduced to vocabularies we think
+# they could belong to (e.g. vcard:includesGroup). We'll make a note of the
+# custom term when used.
+
+vcard:Group {
+  vcard:fn xsd:string {1} ;
+  vcard:note xsd:string ? ;
+  vcard:hasMember vcard:Individual * OR vcard:Organization * ;
```
matthieubosquet, line 18: I'm not familiar with this syntax, does it mean that you can have a mix of 0 to many Individuals and Organizations or 0 to many of one or the other?

shapes/README.md
```diff
@@ -1,106 +1,235 @@
...
+## The 'country' problem?
+An interesting discussion point would be how to model something as seemingly
+trivial as the 'country' field of a user's address. If we take the ISO as just
+an example of a recognized 'authority' when referring to "countries of the
+world" (of course they are by no means the only such authority), then a problem
+arises in that the ISO do not provide IRIs for the countries they otherwise
+standardize identifiers for (e.g., 'ie' for Ireland, 'de' for Germany).
+
+(About 7 years ago I specifically asked the ISO if they might consider minting
+official IRIs for countries. Their reply was, literally, "**_We do not product
```
matthieubosquet, line 22:
```suggestion
official IRIs for countries. Their reply was, literally, "**_We do not produce
```

shapes/README.md
```diff
@@ -1,106 +1,235 @@
...
+see this mailing list thread I started 6 years ago (and specifically see the
+guidance from Martin Hepp, re-iterated in the final email from Bernard Vatant),
+which basically recommends just using the standardized string values for
+countries, and not using IRIs at all(!):
+- Martin Hepp: https://lists.w3.org/Archives/Public/public-vocabs/2015Mar/0165.html
+- Bernard Vatant (last email): https://lists.w3.org/Archives/Public/public-vocabs/2015Mar/0181.html
+
 # Shapes
 
+This remainder of this document describes the overall contents of this
```
matthieubosquet, line 35:
```suggestion
The remainder of this document describes the overall contents of this
```

shapes/README.md
pmcb55: :) That's not a typo (at least not in my post) - that's what the reply from the ISO was, **_literally_**! (I suppose I could add (sic) after it though...)

shapes/contact/group.shex
pmcb55: The latter - the '*' just means 'zero of more', so this is saying the `vcard:hasMember` predicate has the range zero or more `vcard:Individual`, or zero or more `vcard:Organization`.

shapes/README.md
matthieubosquet: ^^, well, that adds salt to the answer

matthieubosquet (commented):
I submitted separate PRs raising concerns about the Address Book entity which I think is essentially a vcard:Group and about the use of a WebID class, which I feel is not appropriate in this context.

I'm also concerned about the idea of caching in index resources. But that would be a separate discussion I believe.

shapes/contact/person.shex
```diff
@@ -22,8 +34,16 @@ vcard:Individual {
...
+    # vcard:AddressBook is a custom class
 }
 
+# We store the preferred WebID for each person. The use of vcard:WebId
+# refer to the Individual's WebIDs, but the idea is that we could expand it
+# to refer to other type of unique identifiers later
+# vcard:WebId is a custom term
+# Big question - SHOULD something as fundamental as WebID be defined in vCard?
 vcard:WebId {
     vcard:value xsd:string {1};
```
matthieubosquet, line 48:
```suggestion
    vcard:value xsd:anyURI * ;
```

Why would a person have one and only one WebID?
I would be very enclined to be more lenient here. And more strict on the type.

As mentioned elsewhere, I don't think we should have a vcard:WebID class and we should have a vcard:hasWebID property.

shapes/contact/person.shex
```diff
@@ -22,8 +34,16 @@ vcard:Individual {
...
+    # vcard:inAddressBook is a custom term
+    # vcard:AddressBook is a custom class
 }
 
+# We store the preferred WebID for each person. The use of vcard:WebId
+# refer to the Individual's WebIDs, but the idea is that we could expand it
+# to refer to other type of unique identifiers later
+# vcard:WebId is a custom term
+# Big question - SHOULD something as fundamental as WebID be defined in vCard?
 vcard:WebId {
```
matthieubosquet, line 47: If we're going to introduce terms to vcard, maybe we should follow their casing convention?

I think that would mean: vcard:WebID instead of vcard:WebId.

pmcb55: Yeah, I definitely free that the natural 'home' for WebID classes or properties is not really vCard (hence my big comment saying 'Big question - ...') on the line above! So I think this addition here in this ShEx is really just a conceptual suggestion at this point, and the details of the name casing would be discussed and decided as part of the bigger WebID discussion.

shapes/contact/person.shex
pmcb55: Yeah, that's certainly a good point (i.e. a Person being limited to just one WebID). I created a separate issue just for that question: https://github.com/inrupt/pod-browser/issues/251

PR #249, opened 2021-02-10
feat: Reference WebIDs in the vCard format

matthieubosquet:
# Reference WebIDs in vCard

It would be great to be able to reference WebIDs in vCard. It would allow to spread them more widely and give them recognition as a thing people use to identify themselves.

Remains the question on which I have close to 0 knowledge. How do we get it through to:
- The vCard Format Specification - https://tools.ietf.org/html/rfc6350
- The W3C ontology - www.w3.org/2006/vcard/ns#

## Deletion

I don't think there is any need for a WebID class and if there was, it would not be defined in vCard. In the end a WebID is just an URI and if standard properties were to be found on WebIDs, it would not be vCard's role to define or maintain them.

On the other hand, the property vcard:hasWebID is definitely something I'd like to see.


pmcb55:
So these comments are suggesting remodelling, which is fine in itself, but was not the intent at all of the original PR. The original PR is merely trying to describe and explain the current modelling - that's all. We can certainly discuss re-modelling this later, but that's a very different PR.

PR #250, opened 2021-02-10
fix: Address books are groups

matthieubosquet:
I don't see a real functional added value in the concept or entity of an AddressBook.

I think they are just Groups.

Why would you only refer to groups from an AddressBook? Why not individuals or organisations...

The range of the property `vcard:hasMember` is `vcard:Kind` which is the parent class of all vcard defined classes: vcard:Group, vcard:Individual, vcard:VCard, vcard:Location, vcard:Organization. It seems more appropriate and flexible to me. It is hard enough to deal with recursion in one place, I don't think we need extra layers of indirection here.


pmcb55:
So these comments are suggesting remodelling, which is fine in itself, but was not the intent at all of the original PR. The original PR is merely trying to describe and explain the current modelling - that's all. We can certainly discuss re-modelling this later, but that's a very different PR.

PR #253, opened 2021-02-11
Feat/multiselect agent picker

VirginiaBalseiro:
# Multiselect agent picker

In this PR:

- adding contacts to agent picker modal table
- changing functionality to allow removing existing agents from a policy (they appeared as checked in the table when they already belong to a policy)
- adding tests
- Agent Picker Modal's filters (search bar and tabs) are functional now

Not in this PR:
- Sharing accordion permissions table tabs are still filtering by "profile.types", since I still do not know what permissions for groups will look like so it will have to be updated once groups permissions are available

# Checklist

### DIFF
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
@@ -99,7 +99,11 @@ export default function AddAgentRow({
       setAgentAvatar(profile.avatar);
       setDisplayedWebId(profile.webId);
     } else if (temporaryRowThing && !profile && contactsArrayLength > 0) {
-      setAgentName(getStringNoLocale(temporaryRowThing, foaf.name) || null);
+      setAgentName(
+        getStringNoLocale(temporaryRowThing, foaf.name) ||
+          getStringNoLocale(temporaryRowThing, vcard.fn) ||
+          null
+      );
       setAgentAvatar(getUrl(temporaryRowThing, vcard.hasPhoto) || null);
       setDisplayedWebId(getUrl(temporaryRowThing, VCARD_WEBID_PREDICATE));
     }
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import PropTypes from "prop-types";
 import { useSession, useThing } from "@inrupt/solid-ui-react";
 import {
@@ -43,6 +43,7 @@ import styles from "./styles";
 import { chain } from "../../../../../src/solidClientHelpers/utils";
 import { fetchProfile } from "../../../../../src/solidClientHelpers/profile";
 import { vcardExtras } from "../../../../../src/addressBook";
+import useContactProfile from "../../../../../src/hooks/useContactProfile";
 
 const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 const VCARD_WEBID_PREDICATE = vcardExtras("WebId");
@@ -89,14 +90,21 @@ export default function AddAgentRow({
   const classes = useStyles();
   const [agentWebId, setAgentWebId] = useState("");
   const [existingPermission, setExistingPermission] = useState();
-
-  const agentName = (thing && getStringNoLocale(thing, foaf.name)) || null;
-  const agentAvatar = (thing && getUrl(thing, vcard.hasPhoto)) || null;
-  const displayedWebId =
-    (thing &&
-      contactsArrayLength > 0 &&
-      getUrl(thing, VCARD_WEBID_PREDICATE)) ||
-    null;
+  const [agentName, setAgentName] = useState(null);
+  const [agentAvatar, setAgentAvatar] = useState(null);
+  const [displayedWebId, setDisplayedWebId] = useState(null);
+  const { data: profile } = useContactProfile(thing);
+  useEffect(() => {
+    if (profile) {
+      setAgentName(profile.name);
+      setAgentAvatar(profile.avatar);
+      setDisplayedWebId(profile.webId);
+    } else if (thing && !profile && contactsArrayLength > 0) {
+      setAgentName(getStringNoLocale(thing, foaf.name) || null);
+      setAgentAvatar(getUrl(thing, vcard.hasPhoto) || null);
+      setDisplayedWebId(getUrl(thing, VCARD_WEBID_PREDICATE));
+    }
+  }, [profile, thing, agentWebId, contactsArrayLength]);
 
   const handleAddAgentsWebIds = async (e) => {
     e.preventDefault();
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
@@ -80,7 +80,6 @@ export default function AddAgentRow({
   newAgentsWebIds,
   contactsArrayLength,
   setAddingWebId,
-  setNoAgentsAlert,
   addingWebId,
   updateTemporaryRowThing,
   permissions,
@@ -113,7 +112,6 @@ export default function AddAgentRow({
       setExistingPermission(true);
       return;
     }
-    setNoAgentsAlert(false);
     setNewAgentsWebIds([...newAgentsWebIds, agentWebId]);
     const newThing = await updateThingForNewRow(
       agentWebId,
@@ -193,7 +191,6 @@ AddAgentRow.propTypes = {
   setNewAgentsWebIds: PropTypes.func.isRequired,
   newAgentsWebIds: PropTypes.arrayOf(PropTypes.string).isRequired,
   setAddingWebId: PropTypes.func.isRequired,
-  setNoAgentsAlert: PropTypes.func.isRequired,
   contactsArrayLength: PropTypes.number.isRequired,
   addingWebId: PropTypes.bool.isRequired,
   updateTemporaryRowThing: PropTypes.func.isRequired,
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
@@ -82,29 +82,29 @@ export default function AddAgentRow({
   setAddingWebId,
   setNoAgentsAlert,
   addingWebId,
-  updateThing,
+  updateTemporaryRowThing,
   permissions,
 }) {
   const { fetch } = useSession();
-  const { thing } = useThing();
+  const { thing: temporaryRowThing } = useThing();
   const classes = useStyles();
   const [agentWebId, setAgentWebId] = useState("");
   const [existingPermission, setExistingPermission] = useState();
   const [agentName, setAgentName] = useState(null);
   const [agentAvatar, setAgentAvatar] = useState(null);
   const [displayedWebId, setDisplayedWebId] = useState(null);
-  const { data: profile } = useContactProfile(thing);
+  const { data: profile } = useContactProfile(temporaryRowThing);
   useEffect(() => {
     if (profile) {
       setAgentName(profile.name);
       setAgentAvatar(profile.avatar);
       setDisplayedWebId(profile.webId);
-    } else if (thing && !profile && contactsArrayLength > 0) {
-      setAgentName(getStringNoLocale(thing, foaf.name) || null);
-      setAgentAvatar(getUrl(thing, vcard.hasPhoto) || null);
-      setDisplayedWebId(getUrl(thing, VCARD_WEBID_PREDICATE));
+    } else if (temporaryRowThing && !profile && contactsArrayLength > 0) {
+      setAgentName(getStringNoLocale(temporaryRowThing, foaf.name) || null);
+      setAgentAvatar(getUrl(temporaryRowThing, vcard.hasPhoto) || null);
+      setDisplayedWebId(getUrl(temporaryRowThing, VCARD_WEBID_PREDICATE));
     }
-  }, [profile, thing, agentWebId, contactsArrayLength]);
+  }, [profile, temporaryRowThing, agentWebId, contactsArrayLength]);
 
   const handleAddAgentsWebIds = async (e) => {
     e.preventDefault();
@@ -115,8 +115,12 @@ export default function AddAgentRow({
     }
     setNoAgentsAlert(false);
     setNewAgentsWebIds([...newAgentsWebIds, agentWebId]);
-    const newThing = await updateThingForNewRow(agentWebId, thing, fetch);
-    updateThing(newThing);
+    const newThing = await updateThingForNewRow(
+      agentWebId,
+      temporaryRowThing,
+      fetch
+    );
+    updateTemporaryRowThing(newThing);
     setAddingWebId(false);
   };
 
@@ -192,6 +196,6 @@ AddAgentRow.propTypes = {
   setNoAgentsAlert: PropTypes.func.isRequired,
   contactsArrayLength: PropTypes.number.isRequired,
   addingWebId: PropTypes.bool.isRequired,
-  updateThing: PropTypes.func.isRequired,
+  updateTemporaryRowThing: PropTypes.func.isRequired,
   permissions: PropTypes.arrayOf(PropTypes.shape()).isRequired,
 };
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
@@ -22,11 +22,14 @@
 import React from "react";
 import userEvent from "@testing-library/user-event";
 import { useSession, useThing } from "@inrupt/solid-ui-react";
-import { createThing } from "@inrupt/solid-client";
+import { addUrl, createThing } from "@inrupt/solid-client";
 import AddAgentRow from "./index";
 import { renderWithTheme } from "../../../../../__testUtils/withTheme";
+import useContactProfile from "../../../../../src/hooks/useContactProfile";
+import { vcardExtras } from "../../../../../src/addressBook";
 
 jest.mock("@inrupt/solid-ui-react");
+jest.mock("../../../../../src/hooks/useContactProfile");
 
 describe("AddAgentRow", () => {
   describe("when adding a new webId", () => {
@@ -34,6 +37,7 @@ describe("AddAgentRow", () => {
     beforeEach(() => {
       useThing.mockReturnValue({ thing: mockThing });
       useSession.mockReturnValue({ fetch: jest.fn() });
+      useContactProfile.mockReturnValue({ data: null });
     });
     const index = 0;
     const setNewAgentsWebIds = jest.fn();
@@ -118,6 +122,67 @@ describe("AddAgentRow", () => {
         "https://somewebid.com/",
         ...newAgentsWebIds,
       ]);
+    });
+  });
+  describe("with existing contacts", () => {
+    beforeEach(() => {
+      useSession.mockReturnValue({ fetch: jest.fn() });
+    });
+    const index = 0;
+    const setNewAgentsWebIds = jest.fn();
+    const newAgentsWebIds = [];
+    const setAddingWebId = jest.fn();
+    const setNoAgentsAlert = jest.fn();
+    const addingWebId = false;
+    const updateThing = jest.fn();
+
+    it("renders a row with the agent's name if profile is available", () => {
+      const mockThing = createThing();
+      useThing.mockReturnValue({ thing: mockThing });
+      useContactProfile.mockReturnValue({
+        data: { webId: "https://somewebid.org", name: "Example", avatar: null },
+      });
+      const { asFragment, getByTestId, queryByText } = renderWithTheme(
+        <AddAgentRow
+          index={index}
+          setNewAgentsWebIds={setNewAgentsWebIds}
+          newAgentsWebIds={newAgentsWebIds}
+          setAddingWebId={setAddingWebId}
+          setNoAgentsAlert={setNoAgentsAlert}
+          addingWebId={addingWebId}
+          updateThing={updateThing}
+          permissions={[]}
+        />
+      );
+      expect(asFragment()).toMatchSnapshot();
+      expect(queryByText("Example")).toBeDefined();
+      expect(getByTestId("agent-webid")).toBeDefined();
+    });
+    it("for newly added webids, if profile is unavailable it renders a row with the provided webId", () => {
+      const mockThing = addUrl(
+        createThing(),
+        vcardExtras("WebId"),
+        "https://somewebid.com"
+      );
+      useThing.mockReturnValue({ thing: mockThing });
+      useContactProfile.mockReturnValue({
+        data: null,
+      });
+      const { asFragment, getByTestId, queryByText } = renderWithTheme(
+        <AddAgentRow
+          index={1}
+          setNewAgentsWebIds={setNewAgentsWebIds}
+          newAgentsWebIds={newAgentsWebIds}
+          setAddingWebId={setAddingWebId}
+          setNoAgentsAlert={setNoAgentsAlert}
+          addingWebId={addingWebId}
+          updateThing={updateThing}
+          permissions={[]}
+        />
+      );
+      expect(asFragment()).toMatchSnapshot();
+      expect(queryByText("https://somewebid.com")).toBeDefined();
+      expect(getByTestId("agent-webid")).toBeDefined();
     });
   });
 });
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
@@ -29,13 +29,14 @@ import useContactProfile from "../../../../../src/hooks/useContactProfile";
 import { vcardExtras } from "../../../../../src/addressBook";
 
 jest.mock("@inrupt/solid-ui-react");
+const mockedThingHook = useThing;
 jest.mock("../../../../../src/hooks/useContactProfile");
 
 describe("AddAgentRow", () => {
   describe("when adding a new webId", () => {
     const mockThing = createThing();
     beforeEach(() => {
-      useThing.mockReturnValue({ thing: mockThing });
+      mockedThingHook.mockReturnValue({ thing: mockThing });
       useSession.mockReturnValue({ fetch: jest.fn() });
       useContactProfile.mockReturnValue({ data: null });
     });
@@ -45,7 +46,7 @@ describe("AddAgentRow", () => {
     const setAddingWebId = jest.fn();
     const setNoAgentsAlert = jest.fn();
     const addingWebId = true;
-    const updateThing = jest.fn();
+    const updateTemporaryRowThing = jest.fn();
     const permissions = [{ webId: "https://example.org/profile/card#me" }];
 
     it("renders a row with an input for the first item in the array", () => {
@@ -57,7 +58,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={[]}
         />
       );
@@ -72,7 +73,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={permissions}
         />
       );
@@ -92,7 +93,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={permissions}
         />
       );
@@ -110,7 +111,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={permissions}
         />
       );
@@ -131,7 +132,7 @@ describe("AddAgentRow", () => {
     const setAddingWebId = jest.fn();
     const setNoAgentsAlert = jest.fn();
     const addingWebId = false;
-    const updateThing = jest.fn();
+    const updateTemporaryRowThing = jest.fn();
 
     beforeEach(() => {
       useSession.mockReturnValue({ fetch: jest.fn() });
@@ -151,7 +152,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={[]}
         />
       );
@@ -177,7 +178,7 @@ describe("AddAgentRow", () => {
           setAddingWebId={setAddingWebId}
           setNoAgentsAlert={setNoAgentsAlert}
           addingWebId={addingWebId}
-          updateThing={updateThing}
+          updateTemporaryRowThing={updateTemporaryRowThing}
           permissions={[]}
         />
       );
--- a/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
+++ b/components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
@@ -125,9 +125,6 @@ describe("AddAgentRow", () => {
     });
   });
   describe("with existing contacts", () => {
-    beforeEach(() => {
-      useSession.mockReturnValue({ fetch: jest.fn() });
-    });
     const index = 0;
     const setNewAgentsWebIds = jest.fn();
     const newAgentsWebIds = [];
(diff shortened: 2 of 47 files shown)


src/addressBook/index.js
```diff
@@ -381,6 +382,32 @@ export function createContact(
   };
 }
 
+// TODO: keeping this separate to maintain functionality but will likely adapt the old function to take a type if we need to reuse it for groups
+export async function findPersonContactInAddressBook(
```
megoth, line 386: Move this into `src/models/contact/person/index.js` - we probably have to copy and rewrite some old functionality into other model modules as well (e.g. we should probably copy an equivalent version of `getProfiles` into `src/models/profile/index.js` - but make sure that it fits the limitations that we have for model modules).

We really should not add more dependencies on this old code module, we want to get rid of it as soon as we can.

src/hooks/useContactProfile/index.js
```diff
@@ -0,0 +1,46 @@
...
+import { useSession } from "@inrupt/solid-ui-react";
+import useSWR from "swr";
+import { asUrl } from "@inrupt/solid-client";
+import { getResource } from "../../solidClientHelpers/resource";
+import { getWebIdUrl } from "../../addressBook";
+import { fetchProfile } from "../../solidClientHelpers/profile";
+
+export default function useContactProfile(contact) {
+  const { fetch } = useSession();
+  let personThingUrl;
```
megoth, line 31: Again, let's move the uses of `let` into specific functions. The use of try catch is also a code smell IMO, but I think it's a viable use in this case. But move it into a separate function to make the code easier to read.

src/hooks/useContactProfile/index.js
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    personThingUrl = asUrl(contact);
+  } catch {
+    personThingUrl = null;
+  }
+  return useSWR(["contact", personThingUrl], async () => {
+    if (!personThingUrl) return null;
+    const {
+      response: { dataset, iri },
+    } = await getResource(personThingUrl, fetch);
```
megoth, line 41: This pattern of loading a contact from the address book, and then load the original profile is something we're doing elsewhere. Let's move that into a reusable function in `src/model/profile/index.js`.

src/models/contact/index.js
```diff
@@ -58,9 +56,7 @@ export async function getContactAll(addressBook, types, fetch) {
     types.map(async (type) => {
       const dataset = await getContactIndexDataset(addressBook, type, fetch);
       return getThingAll(dataset)
-        .filter((contact) =>
-          getUrlAll(contact, rdf.type).includes(type.contactTypeUrl)
-        )
+        .filter((contact) => type.isOfType(contact))
```
megoth, line 59: This needs to be reflected in the test. The current tests are flawed in that the mocked ContactPerson is also adding type `vcard:Individual`, which is wrong. We need to update the mocks to be closer to reality, and update the tests accordingly.

megoth (changes requested):

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
```diff
@@ -89,14 +90,21 @@ export default function AddAgentRow({
...
   const [existingPermission, setExistingPermission] = useState();
-
-  const agentName = (thing && getStringNoLocale(thing, foaf.name)) || null;
-  const agentAvatar = (thing && getUrl(thing, vcard.hasPhoto)) || null;
-  const displayedWebId =
-    (thing &&
-      contactsArrayLength > 0 &&
-      getUrl(thing, VCARD_WEBID_PREDICATE)) ||
-    null;
+  const [agentName, setAgentName] = useState(null);
```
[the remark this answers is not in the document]
VirginiaBalseiro: It is sometimes a contact but it can also be the temporary thing we are adding to the dataset by adding new webId, which is why I didn't go with contact.

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
```diff
@@ -89,14 +90,21 @@ export default function AddAgentRow({
...
+  const [agentAvatar, setAgentAvatar] = useState(null);
+  const [displayedWebId, setDisplayedWebId] = useState(null);
+  const { data: profile } = useContactProfile(thing);
+  useEffect(() => {
+    if (profile) {
+      setAgentName(profile.name);
+      setAgentAvatar(profile.avatar);
+      setDisplayedWebId(profile.webId);
+    } else if (thing && !profile && contactsArrayLength > 0) {
+      setAgentName(getStringNoLocale(thing, foaf.name) || null);
```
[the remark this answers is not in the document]
VirginiaBalseiro: This is in the case there is no profile but there is a thing, which means it's the Thing we created to which we added foaf.name on the fly so we know it's under foaf.name. If it's a contact without a profile it will set the name to null and we display the webId anyway like we do for all failed profiles.

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
```diff
@@ -22,18 +22,22 @@
...
 import { useSession, useThing } from "@inrupt/solid-ui-react";
-import { createThing } from "@inrupt/solid-client";
+import { addUrl, createThing } from "@inrupt/solid-client";
 import AddAgentRow from "./index";
 import { renderWithTheme } from "../../../../../__testUtils/withTheme";
+import useContactProfile from "../../../../../src/hooks/useContactProfile";
+import { vcardExtras } from "../../../../../src/addressBook";
 
 jest.mock("@inrupt/solid-ui-react");
+jest.mock("../../../../../src/hooks/useContactProfile");
```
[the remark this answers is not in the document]
VirginiaBalseiro: No, because we're mocking different resolved values in different describe blocks. In fact I had to put them in separate describe blocks (as opposed to nested describe blocks) because the mocks wouldn't be cleared in between tests.

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
```diff
@@ -109,31 +119,43 @@ export const handleConfirmation = ({
...
     );
-
     if (existingContact.length) {
       return;
     }
 
     if (name) {
       const contact = { webId, fn: name };
-      await saveContact(addressBook, contact, types, fetch);
+      await saveContact(
```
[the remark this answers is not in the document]
VirginiaBalseiro: Yeah, I decided to keep using this one as long as it works until we switch all the code in Contacts instead of modifying this function and breaking the functionality in the Contacts page. I think we should have a separate ticket to do all this work.

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
```diff
@@ -235,23 +262,40 @@ export default function AgentPickerModal({ type, text, onClose }) {
...
+    setContent,
+    setTitle,
     handleSubmitNewWebIds,
   });
 
   const updateThing = (newThing) => {
-    const updatedDataset = setThing(placeholderDataset, newThing);
-    setPlaceholderDataset(updatedDataset);
+    const { dataset: addressBookDataset } = addressBook;
+    const newItem = {
```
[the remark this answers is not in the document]
VirginiaBalseiro: This is for the temporary row that we add to the dataset. It is never saved, it is just to display and empty row in the React SDK table.

src/hooks/useContactProfile/index.js
VirginiaBalseiro: I used it only because I have no other way of finding out if a Thing is local or not, but will remove once that is available on solid client.

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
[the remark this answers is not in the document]
megoth: But the new thing should also represent a contact (albeit a contact in the making)? I guess we could make it explicit, e.g. `contactOrContactToBeCreated`, but that's a bit too verbose. Ok, as long `thing` is easy to understand based on its context, it should be ok.

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.test.jsx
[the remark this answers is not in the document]
megoth: Yes, but we can still say `const mockedThingHook = useThing;` to group the mocked version of `useThing` together with `jest.mock("@inrupt/solid-ui-react");`. You would still use `mockReturnValue` in the various places, but doing `mockedThingHook.mockReturnValue` instead of `useThing.mockReturnValue`.

components/resourceDetails/resourceSharing/addAgentButton/addAgentRow/index.jsx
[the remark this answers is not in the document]
VirginiaBalseiro: It can be a contact, or a contact to be created, or not if it cannot find a profile, so I named it `temporaryRowThing` to make it a bit more readable.

src/addressBook/index.js
```diff
@@ -309,7 +307,7 @@ export async function getIndexDatasetFromAddressBook(
 ) {
   const { respond, error } = createResponder();
   try {
-    const addressBookIri = `${getSourceUrl(addressBookDataset)}#this`; // TODO: Ugly hack, should remove
+    const addressBookIri = `${getSourceIri(addressBookDataset)}#this`; // TODO: Ugly hack, should remove
```
megoth, line 310: AFAICT `getSourceIri` is just an alias for `getSourceUrl`, so I would prefer that we stay consistent in our use of `getSourceUrl`.

src/models/contact/group/index.js
```diff
@@ -53,6 +54,8 @@ export const GROUP_CONTACT = {
   container: GROUP_CONTAINER,
   indexFilePredicate: NAME_GROUP_INDEX_PREDICATE,
   contactTypeUrl: vcard.Group,
+  isOfType: (contact) => getUrlAll(contact, rdf.type).includes(vcard.Group),
```
megoth, line 57: Let's write a test on this

src/models/contact/index.js
```diff
@@ -87,3 +90,44 @@ export async function addContactIndexToAddressBook(addressBook, type, fetch) {
...
+    iri: contactIndexIri,
+  };
+
+  const { response: contacts, error: saveContactsError } = await saveResource(
+    contactResource,
+    fetch
+  );
+
+  if (saveContactsError) return error(saveContactsError);
+  return respond({ iri, contact, contacts });
```
megoth, line 132: I don't want us to use the `{ response, error }` pattern in the model modules - It should return the standardized object or objects of whatever the action has affected.

In this case saving a contact affects the new resource and thing that represents the newly created contact. We also need to ensure that the addressbook exists, and that the people index is exists, as these are created on the fly when needed. That means, as this code stands, it will work when you already have an address book and people index created, but I suspect it won't when that's the case.

Check out `saveGroup` in [src/models/contact/group/index.js](https://github.com/inrupt/pod-browser/blob/main/src/models/contact/group/index.js#L80-L120), that's a good example that shows you what might happen. The corresponding tests also shows what you need to check, namely that it works for a newly created address book that isn't saved in a resource (yet), i.e. an uninstantiated address book, and one that is. You can also see from the that method that it returns a complex object `{ addressBook, group, groupIndex }`, and I suspect we have to do this here as well (but then `{ addressBook, person, personIndex }`).

src/models/contact/person/index.js
```diff
@@ -40,6 +56,9 @@ export const PERSON_CONTACT = {
   indexFile: PEOPLE_INDEX_FILE,
   indexFilePredicate: NAME_EMAIL_INDEX_PREDICATE,
   contactTypeUrl: vcard.Individual,
+  isOfType: (contact) => !!getUrl(contact, vcardExtras("inAddressBook")),
```
megoth, line 59: Let's write a test on this

src/models/profile/index.js
```diff
@@ -0,0 +1,64 @@
...
+      return getResource(url, fetch);
+    })
+  );
+  return profileResponses
+    .filter(({ error }) => !error)
+    .map(({ response }) => response)
+    .map(({ dataset, iri }) => {
+      const thing = getThing(dataset, iri);
+      const avatar = getUrl(thing, vcard.hasPhoto);
+      return addStringNoLocale(getThing(thing, iri), vcard.hasPhoto, avatar);
```
megoth, line 62: This one's a bit strange - we take the URL of a photo, and set is a normal string? Any reason why we need to do it like this?

src/models/contact/person/index.js
```diff
@@ -50,3 +69,85 @@ export function createPersonDatasetUrl(addressBook, id = uuid()) {
...
+  const webIdNode = chain(
+    createThing(),
+    (t) => addUrl(t, rdf.type, vcardExtras("WebId")),
+    (t) => addUrl(t, vcard.value, webId)
+  );
+  const webIdNodeUrl = asUrl(webIdNode, iri);
+  return { webIdNode, webIdNodeUrl };
+}
+
+export async function createPersonContact(addressBookContainerUrl, contact) {
```
megoth, line 93: Seeing this now I realize we need to coordinate a bit with `saveContact` in `src/models/contact/index.js`. Let's have a chat on this.

megoth (changes requested):

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
```diff
@@ -261,63 +295,91 @@ export default function AgentPickerModal({ type, text, onClose }) {
...
+              people: PERSON_CONTACT,
+              groups: GROUP_CONTACT,
+            }}
           />
           <AddWebIdButton onClick={handleAddRow} disabled={addingWebId} />
         </div>
         <AgentsSearchBar handleFilterChange={handleFilterChange} />
-        {contactsArray.length ? (
+        {!contacts && !error && (
+          <Container variant="empty">
```
megoth, line 348: Looks like we should support a variant for the Loading Indicator?

src/addressBook/index.js
```diff
@@ -454,7 +450,7 @@ export async function deleteContact(
   const addressBook = await getSolidDataset(addressBookIri, {
     fetch,
   });
-  const { indexFilePredicate } = TYPE_MAP[type];
+  const { indexFilePredicate } = TYPE_MAP[foaf.Person];
```
megoth, line 453: Shouldn't need to change this I think?

src/models/contact/index.js
```diff
@@ -71,19 +65,27 @@ export async function getContactAll(addressBook, types, fetch) {
...
+      { fetch }
+    );
+    indexUrl = getContactIndexUrl(updatedAddressBook, type);
+  }
+
   return {
-    containerUrl: addressBook.containerUrl,
-    dataset,
-    thing: getThing(dataset, asUrl(addressBook.thing, datasetUrl)),
+    updatedAddressBook,
```
megoth, line 88: Return `updatedAddressBook` as `addressBook`.

src/models/contact/person/index.js
```diff
@@ -50,3 +73,120 @@ export function createPersonDatasetUrl(addressBook, id = uuid()) {
...
+  const dataset = chain(
+    createSolidDataset(),
+    (d) => setThing(d, person),
+    (d) => setThing(d, webIdNode),
+    ...emails
+      .concat(addresses)
+      .concat(telephones)
+      .map(({ thing }) => (d) => setThing(d, thing))
+  );
+  return {
```
megoth, line 142: Is there any reason why we cannot do `savePersonContact` and bundle the save logic here as well? Then the returned object should represent the newly saved person contact, using the pattern `{ dataset, thing }`.

megoth: I see that it might conflict with the method `savePerson`. It looks this might simply be an internal function that allows us to deconstruct the complexity of `savePerson`? I would be ok keeping `createPersonContact` and with returning `{ dataset, iri }` _if_ this is an internal function, i.e. no `export` and not exposing `createContact` on the type object.

src/models/profile/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+export async function getProfileForContact(personContactUrl, fetch) {
+  const {
+    response: { dataset, iri },
+  } = await getResource(personContactUrl, fetch);
+  const webId = getWebIdUrl(dataset, iri);
+  const fetchedProfile = await fetchProfile(webId, fetch);
+  return fetchedProfile;
+}
+export async function getProfilesForPersonContacts(people, fetch) {
```
megoth, line 40: The logic of `getProfileForContact` and `getProfilesForPersonContacts` share a lot of similarities - in fact, I think you could reuse `getProfileForContact` in `getProfilesForPersonContacts`.

megoth (changes requested):

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
megoth: (More like a general question; if you agree, I'll create or add it to a ticket somewhere.)

VirginiaBalseiro: Yes, I just used it a second time so it's a good idea.

src/models/profile/index.js
VirginiaBalseiro: `getProfileForContact` uses `fetchProfile` and so returns the "packaged" Profile whereas the plural one returns the profile Things.

megoth: We should align these the types that these methods return.

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/__snapshots__/index.test.jsx.snap
```diff
@@ -9,7 +9,358 @@ exports[`AgentPickerModal renders a table with tabs, searchbox and an empty stat
     <div
       class="PodBrowser-title"
     >
-      Add Editors for null
+      Edit Editors for null
```
megoth, line 12: Minor nitpick - I think your test is missing something, since we see `null`?

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
```diff
@@ -58,33 +47,72 @@ import AddAgentRow from "../addAgentRow";
...
+import useContacts from "../../../../../src/hooks/useContacts";
+import { GROUP_CONTACT } from "../../../../../src/models/contact/group";
+import {
+  PERSON_CONTACT,
+  findPersonContactInAddressBook,
+  savePerson,
+} from "../../../../../src/models/contact/person";
+import useAddressBook from "../../../../../src/hooks/useAddressBook";
+
+function serializePromises(promiseFactories, postResolveActions) {
```
megoth, line 63: This smells like a reusable function - maybe move it to a utils module? We only have `src/solidClientHelpers/utils.js` till now, but we could introduce a promise model that has these utility functions involving promises.

megoth: Let's move postResolveActions out - Let's avoid callbacks if possible (i.e. just use `await serializePromises` and then list the actions afterwards).

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.test.jsx
```diff
@@ -29,18 +29,24 @@ import AgentPickerModal, {
...
+import mockAddressBook from "../../../../../__testUtils/mockAddressBook";
 import AccessControlContext from "../../../../../src/contexts/accessControlContext";
 import * as ProfileFns from "../../../../../src/solidClientHelpers/profile";
 import ConfirmationDialogContext from "../../../../../src/contexts/confirmationDialogContext";
 import useNamedPolicyPermissions from "../../../../../src/hooks/useNamedPolicyPermissions";
 import usePermissionsWithProfiles from "../../../../../src/hooks/usePermissionsWithProfiles";
+import useAddressBook from "../../../../../src/hooks/useAddressBook";
+import useContacts from "../../../../../src/hooks/useContacts";
 
+jest.mock("../../../../../src/hooks/useAddressBook");
```
megoth, line 46: Let's group the corresponding mocked instances, e.g. `const mockedAddressBookHook = useAddressBook;` in this case.

src/models/contact/index.js
```diff
@@ -71,19 +65,27 @@ export async function getContactAll(addressBook, types, fetch) {
...
-  );
+  let updatedAddressBook = addressBook;
+  let indexUrl = getContactIndexUrl(addressBook, type);
+  if (!indexUrl) {
+    const newIndexUrl = getContactIndexDefaultUrl(
+      addressBook.containerUrl,
+      type
+    );
+    const datasetUrl = getAddressBookIndexUrl(addressBook);
+    updatedAddressBook = await saveSolidDatasetAt(
```
megoth, line 76: Oh, sorry, didn't see this earlier. You're changing the type of `updatedAddressBook` here, from the model `AddressBook` to `SolidDataset`. This won't work.

Make sure to package a new object `{ containerUrl, dataset, thing }` that uses the dataset returned from `saveSolidDatasetAt`:

```
export async function addContactIndexToAddressBook(addressBook, type, fetch) {
  let updatedAddressBook = addressBook;
  let indexUrl = getContactIndexUrl(addressBook, type);
  if (!indexUrl) {
    const newIndexUrl = getContactIndexDefaultUrl(
      addressBook.containerUrl,
      type
    );
    const datasetUrl = getAddressBookIndexUrl(addressBook);
    const updatedAddressBookDataset = await saveSolidDatasetAt(
      datasetUrl,
      setThing(
        addressBook.dataset,
        setUrl(addressBook.thing, type.indexFilePredicate, newIndexUrl)
      ),
      { fetch }
    );
    updatedAddressBook = {
      containerUrl: addressBook.containerUrl,
      dataset: updatedAddressBookDataset,
      thing: getThing(updatedAddressBookDataset, asUrl(addressBook.thing))
    };
    indexUrl = getContactIndexUrl(updatedAddressBook, type);
  }

  return {
    addressBook: updatedAddressBook,
    indexUrl,
  };
}
```

src/models/contact/index.test.js
```diff
@@ -235,13 +235,16 @@ describe("addContactIndexToAddressBook", () => {
   beforeEach(() => {
     mockedSaveSolidDatasetAt = jest
       .spyOn(solidClientFns, "saveSolidDatasetAt")
-      .mockResolvedValue(addressBookWithGroups.dataset);
+      .mockResolvedValue(addressBookWithGroups);
```
megoth, line 238: You probably want to change this back to `addressBookWithGroups.dataset`.

megoth (changes requested):

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/__snapshots__/index.test.jsx.snap
VirginiaBalseiro: The dataset name is fetched from  useContext(DatasetContext)` and I thought it wasn't worth the trouble of trying to mock a context.

megoth: Fair enought

src/solidClientHelpers/utils.js
```diff
@@ -156,3 +156,11 @@ export function sharedStart(...strings) {
   while (i < L && a1.charAt(i) === a2.charAt(i)) i++;
   return a1.substring(0, i);
 }
+
+export async function serializePromises(promiseFactories) {
```
megoth, line 160: Please write a test for this.

megoth (approved):
One final suggestion ^_^

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/addWebIdButton/index.jsx
```diff
@@ -34,15 +34,19 @@ const TESTCAFE_ADD_WEBID_BUTTON = "add-webid-button";
...
   const classes = useStyles();
 
   return (
     <Button
       variant="action"
-      data-testid={TESTCAFE_ADD_WEBID_BUTTON}
-      className={classes.button}
+      data-testid={
+        className && className.includes("mobileOnly")
+          ? "add-webid-button-mobile"
```
megoth, line 46: Let `"add-webid-button-mobile"` be set to a constant as well.

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/addWebIdButton/index.jsx
```diff
@@ -34,15 +34,19 @@ const TESTCAFE_ADD_WEBID_BUTTON = "add-webid-button";
...
+          : TESTCAFE_ADD_WEBID_BUTTON
+      }
+      className={clsx(classes.button, className && className)}
```
megoth, line 49: Think you can do just `className` here

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/addWebIdButton/styles.js
```diff
@@ -24,6 +24,11 @@ import { createStyles } from "@solid/lit-prism-patterns";
 export default function styles(theme) {
   return createStyles(theme, ["icons", "button"], {
     button: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 27: I prefer to do mobile first for CSS, i.e. let the rules for smaller screen come first, and then add breakpoints for larger widths. Please flip `theme.breakpoints.down` into `theme.breakpoints.up`. (Not 100% necessary if it's hard, but would like to stay consistent.)

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/addWebIdButton/index.jsx
```diff
@@ -34,15 +34,19 @@ const TESTCAFE_ADD_WEBID_BUTTON = "add-webid-button";
...
   const bem = useBem(useStyles());
   const classes = useStyles();
 
   return (
     <Button
       variant="action"
-      data-testid={TESTCAFE_ADD_WEBID_BUTTON}
-      className={classes.button}
+      data-testid={
+        className && className.includes("mobileOnly")
```
megoth, line 45: Let's do `className?.includes("mobileOnly")` instead?

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/index.jsx
```diff
@@ -355,9 +356,21 @@ export default function AgentPickerModal({ type, text, onClose, setLoading }) {
               groups: GROUP_CONTACT,
             }}
           />
-          <AddWebIdButton onClick={handleAddRow} disabled={addingWebId} />
+          <MobileAgentsSearchBar handleFilterChange={handleFilterChange} />
+
+          <AddWebIdButton
```
megoth, line 361: Perhaps more readable/easier to work with to use the Hidden component in Material UI? https://material-ui.com/components/hidden/

megoth: Means you don't need to introduce CSS classes that does the same (i.e. `mobileOnly` and `desktopOnly`).

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/styles.js
```diff
@@ -24,10 +24,18 @@ import { createStyles } from "@solid/lit-prism-patterns";
 const MODAL_MIN_WIDTH = 680;
 const MODAL_MIN_HEIGHT = 680;
 const TABLE_MAX_HEIGHT = 460;
+const MOBILE_TABLE_MIN_HEIGHT = 300;
 
 export default function styles(theme) {
   return createStyles(theme, ["icons", "button", "table"], {
     paper: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 32: Would like to flip to `theme.breakpoints.up` here as well

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/styles.js
```diff
@@ -45,6 +53,9 @@ export default function styles(theme) {
       marginBottom: theme.spacing(1.2),
     },
     tableContainer: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 56: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/styles.js
```diff
@@ -97,6 +108,15 @@ export default function styles(theme) {
       alignItems: "center",
     },
     buttonsContainer: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 111: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/styles.js
```diff
@@ -105,12 +125,20 @@ export default function styles(theme) {
       justifyContent: "space-between",
     },
     cancelButton: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 128: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/addAgentButton/agentPickerModal/styles.js
```diff
@@ -105,12 +125,20 @@ export default function styles(theme) {
...
+        textAlign: "center",
+        width: "100%",
+      },
       fontSize: theme.typography.body1.fontSize,
       fontWeight: theme.typography.body1.fontWeight,
       padding: theme.spacing(1.4, 1.8),
       backgroundColor: "transparent",
     },
     submitAgentsButton: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 138: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/agentsSearchBar/styles.js
```diff
@@ -24,6 +24,9 @@ import { createStyles } from "@solid/lit-prism-patterns";
 export default function styles(theme) {
   return createStyles(theme, ["icons", "table"], {
     searchBoxContainer: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 27: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/styles.js
```diff
@@ -0,0 +1,82 @@
...
+
+export default function styles(theme) {
+  return createStyles(theme, ["icons", "table"], {
+    searchBoxContainer: {
+      display: "none",
+      margin: theme.spacing(0.7, 1.2),
+      border: `1px solid ${theme.palette.grey.A100}`,
+      borderRadius: "10px",
+      height: "2.5rem",
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 34: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/styles.js
```diff
@@ -0,0 +1,82 @@
...
+          right: theme.spacing(0.25),
+          left: theme.spacing(0.25),
+        },
+        "&.hidden": {
+          width: "min-content",
+        },
+      },
+    },
+    searchInput: {
+      [theme.breakpoints.down("xs")]: {
```
megoth, line 54: And here (`theme.breakpoints.up`)

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/styles.js
```diff
@@ -0,0 +1,82 @@
...
+    searchBoxContainer: {
+      display: "none",
+      margin: theme.spacing(0.7, 1.2),
+      border: `1px solid ${theme.palette.grey.A100}`,
+      borderRadius: "10px",
+      height: "2.5rem",
+      [theme.breakpoints.down("xs")]: {
+        border: "none",
+        display: "flex",
+        "&.expanded": {
```
megoth, line 37: Please use the modifier pattern in BEM (so should be `searchBoxContainer--expanded`). You can reflect this easily with `bem("searchBoxContainer", "expanded")` (or `bem("searchBoxContainer", { expanded: trueOrFalse })`).

megoth (changes requested):

megoth:
Just adding a bit more feedback from testing it locally:

I reckon you might want to set a maxWidth for the input when adding a WebID?

bilde

We probably want to adjust the prominent buttons for smaller screens (Save viewers and Cancel is very big here); Nothing you need to mind, I think, but will ping @KyraAssaad about it:

bilde

Nitpick: Would be nice to get focus on the search field when opening search field on mobile

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+  useEffect(() => {
+    if (expanded) {
+      textInput.current.focus();
+    }
+  }, [expanded, textInput]);
+
+  return (
+    <div
+      className={
+        (bem("searchBoxContainer"),
```
megoth, line 53: First, I was confused by the syntax of this (`className={(expression, expression)}`), second, I think it's unnecessary. The second expression (`bem("searchBoxContainer", expanded ? "expanded" : "hidden")`) will result in `searchBoxContainer searchBoxContainer--expanded` or `searchBoxContainer searchBoxContainer--hidden`.

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/styles.js
```diff
@@ -0,0 +1,90 @@
...
+      fontSize: "0.8125rem",
+      width: "100%",
+      font: "inherit",
+      fontWeight: 500,
+      color: theme.palette.primary.text,
+    },
+    iconSearch: {
+      fontSize: theme.typography.body1.fontSize,
+    },
+    iconCloseExpanded: {
```
megoth, line 77: I would prefer that we stay consistent with the BEM style (i.e. `icon-close--expanded`).

megoth: But I'll leave it up to you to decide if it's worth changing at this point.

megoth (approved):
Only smaller stuff remaining, which is not critical.

components/resourceDetails/resourceSharing/mobileAgentsSearchBar/styles.js
VirginiaBalseiro: I need to use the classes like this to override the `root` class from MUI

kevin-inrupt (approved):

PR #255, opened 2021-02-19
Update buttons

megoth:
# Updating UI for buttons

Updated primary buttons or switched them out with apprioriate variants.

Note: Haven't fixed Edit profile page completely, as it includes somewhat more complex component configuration than what we'll easily fix with the Prism buttons. But we'll work on profile page in near future, so we'll leave it for then.

Is dependent on https://github.com/inrupt/prism/pull/134, so will build a proper version once that gets through the pipeline.

# Checklist


ajacksified (commented):
Looks good - you'll also want to update the "delete" button in the resource menu.

megoth:
> Looks good - you'll also want to update the "delete" button in the resource menu.

You mean what is currently shown as a red link today? 

bilde

We don't have a dangerous button like this yet, so didn't want to switch it out. Also, for non-container resources you can see a "Download"-button, and we might want to distinguish this "list of actions" a bit differently than a list of buttons.

@KyraAssaad What do you think, should we switch out this button?

ajacksified:
> bilde

Also: the left-padding on the close button.

megoth:
> > bilde
> 
> Also: the left-padding on the close button.

Meaning we should do a negative margin left (to make sure that the states that require that padding are shown properly)?

bilde

It's fairly easy to do: 

bilde

But it means that the non-default states look a bit funky:

bilde

Another alternative is to present the close button as a small button, which gives it a slight background color:

bilde

But that might be a tad to prominent?

megoth:
> > bilde
> 
> Also: the left-padding on the close button.

Ok, after discussion with @ajacksified and @KyraAssaad in other channels, we landed on displaying the close button to the right:

bilde

Requires the work in https://github.com/inrupt/prism/pull/135.

ajacksified (approved):

PR #269, opened 2021-03-22
Feat: displaying inherited access

VirginiaBalseiro:
# Displaying inherited access 

- add function to get inherited permissions

- modify getPermissions functions so that they return all the permissions that apply to the resource, including inherited ones

- add tests

# Checklist

### DIFF
--- a/src/accessControl/acp/index.js
+++ b/src/accessControl/acp/index.js
@@ -49,7 +49,7 @@ import {
 import { isHTTPError } from "../../error";
 import { PUBLIC_AGENT_PREDICATE } from "../../models/contact/public";
 import { AUTHENTICATED_AGENT_PREDICATE } from "../../models/contact/authenticated";
-import { namedPolicies } from "../../../constants/policies";
+import { namedPolicies, POLICIES_TYPE_MAP } from "../../../constants/policies";
 
 export const noAcrAccessError =
   "No access to Access Control Resource for this resource";
@@ -75,19 +75,19 @@ const acpMapForCustomApplyPolicies = {
 export const getPolicyNameFromAccess = (access) => {
   const { read, write, append } = access;
   if (read && write && !append) {
-    return "editors";
+    return POLICIES_TYPE_MAP.editors.name;
   }
   if (read && !write && !append) {
-    return "viewers";
+    return POLICIES_TYPE_MAP.viewers.name;
   }
   if (read && append && !write) {
-    return "viewAndAdd";
+    return POLICIES_TYPE_MAP.viewAndAdd.name;
   }
   if (append && !write && !write) {
-    return "addOnly";
+    return POLICIES_TYPE_MAP.addOnly.name;
   }
   if (write && !append && !read) {
-    return "editOnly";
+    return POLICIES_TYPE_MAP.editOnly.name;
   }
   return null;
 };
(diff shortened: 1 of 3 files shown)


src/accessControl/acp/index.js
```diff
@@ -385,14 +417,92 @@ export default class AcpAccessControlStrategy {
...
+      const containerUrl = getPolicyResourceUrl(
+        this.#originalWithAcr,
+        this.#policiesContainerUrl,
+        name
+      );
+      return getNamedPolicyUrl(containerUrl, name);
+    });
+    const inheritedPoliciesUrls = policies.filter(
+      (policyUrl) =>
+        !namedPolicyUrls.includes(policyUrl) && !policyUrl.includes("#Owner")
```
megoth, line 434: This smells a bit hard-coded, and could potensially lead to problems, e.g. what if someone else creates a policy that starts with `Owner`? It's an edge-case, and probably not going to happen, but is there another way of inferring (and consequently exclude) Owner policies?

src/accessControl/acp/index.js
```diff
@@ -385,14 +417,92 @@ export default class AcpAccessControlStrategy {
...
+            fetch: this.#fetch,
+          });
+          const modesAndAgentsForUrl = getNamedPolicyModesAndAgents(
+            url,
+            dataset
+          );
+          return modesAndAgentsForUrl;
+        })
+      );
+      const permissions = [];
```
megoth, line 455: You declare `permissions` as an array, but treat it as an object (map) later on. Let's change this to `const permissions = {};`

megoth (approved):
A minor change needed, but otherwise looks good

src/accessControl/acp/index.js
ajacksified: Also, "#Owner" should be a constant.

src/accessControl/acp/index.js
```diff
@@ -63,6 +67,26 @@ const acpMapForCustomApplyPolicies = {
   addOnly: createAcpMap(false, false, true),
 };
 
+export const getPolicyNameFromAccess = (access) => {
+  const { read, write, append } = access;
+  if (read && write && !append) {
+    return "editors";
```
ajacksified, line 73: All strings should be exported constants so that they can be used easily in tests, and so that if the string values are used multiple times they all come from a single definition.

src/accessControl/acp/index.js
VirginiaBalseiro: After a discussion with @megoth where we considered different options, we decided to go with filtering based on the policies container url. If/when we switch from `pb_policies` to `policies`, the owner will no longer be filtered out, so we have updated the corresponding ticket so we can keep this in mind when making that change.

src/accessControl/acp/index.js
```diff
@@ -385,14 +417,94 @@ export default class AcpAccessControlStrategy {
...
+        this.#originalWithAcr,
+        this.#policiesContainerUrl,
+        name
+      );
+      return getNamedPolicyUrl(containerUrl, name);
+    });
+    const inheritedPoliciesUrls = policies.filter(
+      (policyUrl) =>
+        !namedPolicyUrls.includes(policyUrl) &&
+        policyUrl.startsWith(this.#policiesContainerUrl)
```
megoth, line 435: This is inline with what we discussed - hopefully we can find a better approach later, but this should work for now

ajacksified (changes requested):

ajacksified (approved):

PR #281, opened 2021-04-07
Fix: adding tests to improve coverage

VirginiaBalseiro:
This PRs adds tests for:

- `addFileButton`
- `profile`

in order to improve test coverage.

### DIFF
--- a/components/addFileButton/index.test.jsx
+++ b/components/addFileButton/index.test.jsx
@@ -181,6 +181,30 @@ describe("handleSaveResource", () => {
     expect(setAlertOpen).toHaveBeenCalledWith(true);
   });
 
+  test("it returns a handler that gives feedback to user if an error occurs", async () => {
+    const fileName = "myfile.txt";
+    const newFilePath = currentUri + encodeURIComponent(fileName);
+
+    jest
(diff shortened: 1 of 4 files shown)


components/addFileButton/index.test.jsx
```diff
@@ -231,15 +255,24 @@ describe("handleFileSelect", () => {
   });
 
   test("it returns a handler that uploads a file", async () => {
```
ajacksified, line 257: Test no longer should be async (as well as the changed tests below)

ajacksified (changes requested):

components/addFileButton/index.test.jsx
```diff
@@ -231,15 +255,24 @@ describe("handleFileSelect", () => {
...
-    await handler({ target: { files: [file] } });
+    handler({ target: { files: [file] } });
```
megoth, line 258: A bit skeptic to the handler _not_ being asynchronous, but the code checks out. Just to clean up a bit more, you can remove the async from the test as well.

megoth (approved):

ajacksified (approved):

PR #282, opened 2021-04-08
Fix/requesting dataset twice

VirginiaBalseiro:
This PR fixes bug where every resource was requested twice. When navigating to a container, two GET requests were made for the container.

The bug was caused by a hook that fetched the IRIs for the resources within a container. Instead of using the dataset directly, it was fetching the dataset again, and then fetching the IRIs. 

Since we already do a GET request for the dataset itself before fetching the IRIs for the resources, it made more sense to use the dataset directly. In addition, we also used the container Thing to get the IRIs, so I created a container model and we now use the model in the Container view component and the function that fetches the IRIs. 

**Update**:

After looking into it, I could not get the `mutate` function from `useSWR` to behave in the way it is supposed to behave (or the way I would expect it to behave). The only way to get `mutate` to update the data properly was to pass it the newly fetched data, which kind of beats the purpose since what I wanted was to remove the extra request for the dataset. I decided to rewrite the hook to simply use a `useEffect`. 

I also introduced a spinner when moving from one folder to another since the data takes a bit longer than the breadcrumbs to update and it can be disconcerting. 

# To test:
- navigate to a container
- in the Network tab, verify that only one GET request is made when navigating to the container


components/container/index.test.jsx
```diff
@@ -144,7 +146,7 @@ describe("Container view", () => {
   });
 
   it("renders Not supported if the fetch for container returns 500", () => {
-    mockedDatasetHook.mockReturnValue({
+    mockedContainerHook.mockReturnValue({
       error: { message: "500" },
```
megoth, line 150: Let's improve this slightly by doing `error: new Error("500")` instead.

megoth: Same with the other cases of `{ message: --- }`

components/container/index.test.jsx
```diff
@@ -156,8 +158,9 @@ describe("Container view", () => {
 
   it("renders Not Supported if resource loaded is not a container", () => {
     const resourceIri = "http://example.com/resource";
-    mockedDatasetHook.mockReturnValue({
-      data: mockSolidDatasetFrom(resourceIri),
+    mockedContainerHook.mockReturnValue({
+      data: { dataset: mockSolidDatasetFrom(resourceIri) },
```
megoth, line 162: You're creating very specific objects that only has the properties you need for this test - you should instead create mocked objects of models (with functions that can be reused elsewhere).

For example, we could have a factory method that takes the URL of the thing you want, which mocks a thing and dataset (which you must strip the hash-part of the URL, if any) for you based on that one URL.

megoth (changes requested):

src/models/container/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+/* Model functions */
+
+export async function getContainer(url, { fetch }) {
+  const dataset = await getSolidDataset(url, { fetch });
+  const thing = getThing(dataset, url);
+  return { dataset, thing };
+}
+
+export function getContainerResourceIris({ thing }) {
```
megoth, line 33: Let's follow the same naming convention as the Solid Client library, i.e. `getContainerResourceUrlAll`

megoth (approved):

ajacksified (approved):
Nice work!

kevin-inrupt (approved):

PR #292, opened 2021-05-03
Chore: upgrade solid-ui-react

VirginiaBalseiro:
- upgrade solid-ui-react to latest version

- fix breaking changes: DatasetProvider and SessionProvider

- update tests


components/profile/index.test.jsx
```diff
@@ -33,10 +40,19 @@ describe("Profile", () => {
...
+      const { asFragment } = renderWithTheme(
+        <SessionProvider>
+          <Profile profileIri={profileIri} editing />
+        </SessionProvider>
+      );
+      expect(asFragment()).toMatchSnapshot();
+    });
+    expect(screen.findByTestId(TESTCAFE_ID_NAME_FIELD)).not.toBeNull();
+    expect(screen.findByTestId(TESTCAFE_ID_ROLE_FIELD)).not.toBeNull();
+    expect(screen.findByTestId(TESTCAFE_ID_ORG_FIELD)).not.toBeNull();
```
Vinnl, line 55: I don't quite have the context to fix this entirely, but if you try to make the code invalid, this test will still succeed. The `findBy*` functions return a Promise that either resolve successfully or reject, so if you want to test whether an element is present you should test for the Promise to resolve. This would require making the test function `async` and then doing [something like](https://jestjs.io/docs/asynchronous#resolves--rejects):

```suggestion
    await expect(screen.findByTestId(TESTCAFE_ID_NAME_FIELD)).resolves.not.toBeNull();
    await expect(screen.findByTestId(TESTCAFE_ID_ROLE_FIELD)).resolves.not.toBeNull();
    await expect(screen.findByTestId(TESTCAFE_ID_ORG_FIELD)).resolves.not.toBeNull();
```

Alternatively, it looks like the `queryBy*` functions do do what you expect:

```suggestion
    expect(screen.queryByTestId(TESTCAFE_ID_NAME_FIELD)).not.toBeNull();
    expect(screen.queryByTestId(TESTCAFE_ID_ROLE_FIELD)).not.toBeNull();
    expect(screen.queryByTestId(TESTCAFE_ID_ORG_FIELD)).not.toBeNull();
```

Unfortunately, when I apply the above, the tests start failing, but you're probably in a better position to fix that than I am.

src/hooks/useAccessControl/index.test.jsx
```diff
@@ -76,7 +81,7 @@ describe("useAccessControl", () => {
     await waitForNextUpdate();
     expect(result.current.accessControl).not.toBeNull();
     rerender(resourceInfo + 2);
-    expect(result.current.accessControl).toBeNull();
+    expect(result.all[0].accessControl).toBeNull(); // checking that the first render sets accessControl to null
```
Vinnl, line 84: I don't really understand why this change is necessary - why shouldn't the last render set it to null? (Ideally, the comment describes the why, rather than the what.)

src/models/contact/group/index.test.js
```diff
@@ -99,6 +100,18 @@ const groupIndexWithGroup1And2Dataset = chain(
       group2Url
     )
 );
+const group1Thing = getThing(groupIndexWithGroup1And2Dataset, group1Url);
+const groupIndexAfterRemovingGroup1 = chain(
+  groupIndexWithGroup1And2Dataset,
+  (d) =>
+    removeGroupFromMockedIndexDataset(
```
Vinnl, line 107: I'm probably lacking a bit too much context here - could you maybe clarify how the changes in solid-ui-react require the changes in this file, as well as the change in the addressBook's test in which `3` suddenly had to be replaced by `4`, which I suppose is related?

Vinnl (commented):
The things I understand looks good, but there are a couple of things I don't understand yet :) Also the new test of the Profile component doesn't actually test anything as far as I can see.

components/profile/index.test.jsx
VirginiaBalseiro: Great catch! This actually helped me find a bug I had missed 🐛

src/hooks/useAccessControl/index.test.jsx
VirginiaBalseiro: I found that `current` checks the current render, which is actually after the data is available and returned, so I had to checked what happened before the data was requested and available, since the test is to check that it is set to null while loading.

src/models/contact/group/index.test.js
VirginiaBalseiro: I found that a lot of the order in which the datasets/things extracted from datasets were being returned changed. I assumed this could be related to `solid-client` but I don't know for sure.

Vinnl: Hmm, weird. I'm not sure what could cause this, but since the order doesn't matter for the actual code, probably not worth spending any time on diving into that.

Vinnl (approved):
Good to hear you found and fixed another bug. I can't really see from the code what the bug was (and whether it makes sense to add a test to catch potential regressions), but the implementation looks good to me.

VirginiaBalseiro:
> Good to hear you found and fixed another bug. I can't really see from the code what the bug was (and whether it makes sense to add a test to catch potential regressions), but the implementation looks good to me.

The tests *actually failing* instead of giving a false positive were the catch :) the fields weren't rendered because the dataset wasn't available (I had to switch to `solidDataset` for the return value of the DatasetContext as well)

kevin-inrupt (approved):
All TestCafe tests pass against this PR.

PR #293, opened 2021-05-07
Feat/use latest acp

VirginiaBalseiro:
In this PR:

- using the latest ACP (v3)

### DIFF
--- a/src/accessControl/acp/index.js
+++ b/src/accessControl/acp/index.js
@@ -20,8 +20,7 @@
  */
 
 import {
-  acp_v1 as acpv1,
-  acp_v2 as acpv2,
+  acp_v3 as acp,
   asUrl,
   deleteFile,
   getSolidDataset,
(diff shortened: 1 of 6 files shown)


kevin-inrupt (approved):
All TestCafe tests pass against this PR.

src/accessControl/acp/index.js
```diff
@@ -323,76 +330,49 @@ function getPolicyUrlName(policyUrl, name) {
...
-  } catch (error) {
-    // TODO: Handle this error (probably by replacing acpv1.setControl with newer ACP APIs)
-    // Same problem as noted above, but this time it's acpv1.setControl that fails to handle
-    // datasetWithAcr properly. It doesn't seem to affect the outcome though, so we'll leave
-    // it like this for now
-  }
+  const acr = acp.addPolicyUrl(datasetWithAcr, policyUrl);
   return {
-    changed: changed || !existingAccessControl,
+    changed: changed || !existingPolicies,
```
Vinnl, line 375: Given that you've already returned above if `existingPolicies` is truthy, isn't this always true?

```suggestion
    changed: true,
```

src/accessControl/acp/index.test.js
```diff
@@ -380,118 +401,112 @@ describe("AcpAccessControlStrategy", () => {
...
       );
-      expect(acpFns2.saveAcrFor).toHaveBeenCalledWith(
+      expect(acpFns3.saveAcrFor).toHaveBeenCalledWith(
         datasetWithAcrControlAccess,
         {
           fetch,
         }
       );
     });
     it("deletes rule and policy is agent is the last one on the list", async () => {
```
Vinnl, line 466: Heh, sorry for going outside of the scope of this PR, but typo was such an easy fix that I couldn't resist :)

```suggestion
    it("deletes rule and policy if agent is the last one on the list", async () => {
```

Vinnl (approved):
Looks like it was mostly relatively straightforward renames - hope the migration wasn't too painful.

PR #296, opened 2021-06-01
Fix: pass restorePreviousSession to SessionProvider 

VirginiaBalseiro:
In this PR:

- upgrade to latest version of `solid-ui-react`
- pass `restorePreviousSession` to `SessionProvider` to restore session on browser refresh
- use `onSessionRestore` to redirect to navigated page

This PR fixes bug #SDK-2048.

### DIFF
--- a/pages/_app.jsx
+++ b/pages/_app.jsx
@@ -27,14 +27,14 @@ import { MatomoProvider, createInstance } from "@datapunt/matomo-tracker-react";
 
 import PropTypes from "prop-types";
 import Head from "next/head";
-import { useRouter } from "next/router";
+import router, { useRouter } from "next/router";
 import CssBaseline from "@material-ui/core/CssBaseline";
 
 import { create } from "jss";
 import preset from "jss-preset-default";
 
 import { appLayout, useBem } from "@solid/lit-prism-patterns";
-import { SessionProvider } from "@inrupt/solid-ui-react";
+import { SessionProvider, useSession } from "@inrupt/solid-ui-react";
 import {
   createStyles,
   makeStyles,
@@ -136,7 +136,11 @@ export default function App(props) {
       <MatomoProvider value={matomoInstance}>
         <StylesProvider jss={jss}>
           <ThemeProvider theme={theme}>
-            <SessionProvider sessionId="pod-browser" restorePreviousSession>
+            <SessionProvider
+              sessionId="pod-browser"
+              restorePreviousSession
+              onSessionRestore={(url) => router.push(url)}
+            >
               <FeatureProvider>
                 <AlertProvider>
                   <ConfirmationDialogProvider>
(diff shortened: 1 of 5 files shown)


pages/_app.jsx
```diff
@@ -136,7 +136,7 @@ export default function App(props) {
       <MatomoProvider value={matomoInstance}>
         <StylesProvider jss={jss}>
           <ThemeProvider theme={theme}>
-            <SessionProvider sessionId="pod-browser">
+            <SessionProvider sessionId="pod-browser" restorePreviousSession>
```
Vinnl, line 139: This flag alone unfortunately isn't enough (otherwise it'd be enabled by default) - you still have to make sure PodBrowser actually navigates back to the page the user was at before the session was revalidated. See https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/restore-session-browser-refresh/#use-session-restore-event-handler

(To reproduce what goes wrong if you don't do this: sign in, navigate to a different page, then refresh. After the session is restored, you will be back at the page you were at when you signed in, rather than where you were when you refreshed.)

pages/_app.jsx
```diff
@@ -27,14 +27,14 @@ import { MatomoProvider, createInstance } from "@datapunt/matomo-tracker-react";
...
-import { useRouter } from "next/router";
+import router, { useRouter } from "next/router";
 import CssBaseline from "@material-ui/core/CssBaseline";
 
 import { create } from "jss";
 import preset from "jss-preset-default";
 
 import { appLayout, useBem } from "@solid/lit-prism-patterns";
-import { SessionProvider } from "@inrupt/solid-ui-react";
+import { SessionProvider, useSession } from "@inrupt/solid-ui-react";
```
Vinnl, line 37: Heads up: some linting warnings here about an unused import:

```suggestion
import { SessionProvider } from "@inrupt/solid-ui-react";
```

pages/_app.jsx
```diff
@@ -27,14 +27,14 @@ import { MatomoProvider, createInstance } from "@datapunt/matomo-tracker-react";
 
 import PropTypes from "prop-types";
 import Head from "next/head";
-import { useRouter } from "next/router";
+import router, { useRouter } from "next/router";
```
Vinnl, line 30: Heh, this is interesting. This seems to work, but I can't find anything in [the docs](https://nextjs.org/docs/api-reference/next/router) about a default export from `next/router`. I think to be sure it's probably safer to align with the docs and call `.push` on the return value of `useRouter`.

Vinnl (approved):
:+1: verified working. I'd slightly change it to the way Next.js appears to document how the router works, unless I've missed the relevant docs, just in case this is using a deprecated API or something that they might remove later.

pages/_app.jsx
VirginiaBalseiro: Oh that's so weird. Probably auto-imported and I didn't notice.

kevin-inrupt (approved):

PR #299, opened 2021-06-09
Fix: remove unnecessary dependency in hook causing rerenders

### LINKED ISSUE #294: User interface blinking when trying to manage a file
Hello !

When I want to manage a file, the interface on the right is blinking, sometimes so much than I can't do anything.

It occurs on the page `https://podbrowser.inrupt.com/resource/` with `?action=details`.

VirginiaBalseiro:
This PR removes the dependency on `authenticatedProfile` in the `useAccessControl` hook, since it was not used and it was causing the details drawer to rerender (causing the blinking effect described in the issue).

It is unclear to me why this was added to the `useEffect`  dependencies, but it did not seem to cause any issues to remove it. 

This PR fixes bug #SDK-2050 and fixes #294.

### DIFF
--- a/src/hooks/useAccessControl/index.jsx
+++ b/src/hooks/useAccessControl/index.jsx
@@ -24,11 +24,9 @@ import { useSession } from "@inrupt/solid-ui-react";
 import { getSourceUrl } from "@inrupt/solid-client";
 import { getAccessControl, isAcp } from "../../accessControl";
 import usePoliciesContainerUrl from "../usePoliciesContainerUrl";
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
As identified by @acailly at https://github.com/inrupt/pod-browser/issues/294#issuecomment-855978215 - thanks! :clap: 

Since #243, identified by `git blame` as the commit that introduced it, did not mention anything about this, I too expect this to be an oversight. It's weird that the React Hooks ESLint rules didn't catch this though :/

kevin-inrupt (approved):

PR #302, opened 2021-06-17
Fix: adding check for confirmation dialog

VirginiaBalseiro:
In this PR:

- checking for confirmation dialog id before calling savePermissions for user changing their own permissions

We weren't checking if the confirmation dialog being confirmed was the permissions change dialog, so we were calling `savePermissions` every time any dialog was confirmed, which was causing a change in permissions before the folder was deleted. 

The confirmation dialog is particularly hard to test which is why we didn't catch this before, and we have a ticket to refactor it (using a simple dialog component instead of context wherever we need it). 

This PR fixes bug #SDK-2085.


Vinnl (approved):
Oh wow, this one looks like a major pain in the underside to track down, good job.

kevin-inrupt (approved):

PR #304, opened 2021-07-06
Feature/app authentication

norbert989:
# Leverage App Authentication

This PR adds app authentication for PodBrowser via a clientID file generated on the `api/app` route depending on the `IDP` used with a fallback to dynamic registration if `oidc` is not supported or a fallback to the public oidc client for local dev `http://www.w3.org/ns/solid/terms#PublicOidcClient`

# Checklist


public/app.jsonld
```diff
@@ -0,0 +1,9 @@
+{
+    "@context": "https://www.w3.org/ns/solid/oidc-context.jsonld",
+
+    "client_id": "https://pod-browser-git-feature-app-authentication-inrupt.vercel.app/app.jsonld#id",
```
kay-kim, line 4: no hash

public/app.jsonld
```diff
@@ -0,0 +1,9 @@
...
+    "redirect_uris": ["https://pod-browser-git-feature-app-authentication-inrupt.vercel.app/", "https://localhost:3000/"],
```
kay-kim, line 5: Your redirect_uris value needs to match what you pass to the login()

components/login/provider/index.jsx
```diff
@@ -46,6 +46,7 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 const providers = getIdentityProviders();
 export const TESTCAFE_ID_LOGIN_FIELD = "login-field";
 const TESTCAFE_ID_GO_BUTTON = "go-button";
+const CLIENT_APP_WEBID = "https://podbrowser.inrupt.com/app.jsonld#id";
```
kay-kim, line 49: I'd probably remove the hash since the whole file acts as the identifier file

ajacksified (changes requested):

pages/api/app.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+function buildAppProfile(hostname, clientId) {
+  return {
+    "@context": "https://www.w3.org/ns/solid/oidc-context.jsonld",
+    client_id: clientId,
+    redirect_uris: [hostname],
+    client_name: "Inrupt PodBrowser",
+    grant_types: ["refresh_token", "authorization_code"],
+    response_types: ["code"],
```
kay-kim, line 29: So ... per discussion with Aaron, the default values for grant_types and response_types should be correct.  So, whether you all want to specify grant_types and response_types, I'll leave it up to you all.  (I did ask Aaron if you all should.  When he gets back, I'll forward it on.  If he states otherwise, I'll let you all know).

kay-kim: Per Aaron -- he said it's up to you all

> That's really up to them. There is a strong argument to explicitly defining the default values *in code*

components/login/index.jsx
```diff
@@ -27,17 +27,21 @@ import { useBem } from "@solid/lit-prism-patterns";
...
 const TESTCAFE_ID_LOGIN_TITLE = "login-title";
 export const TESTCAFE_ID_OTHER_PROVIDERS_BUTTON = "other-providers-button";
 const PROVIDER_IRI = "https://broker.pod.inrupt.com/";
-
-const authOptions = {
-  clientName: "Inrupt PodBrowser",
-};
+const hostname = getCurrentHostname();
+const CLIENT_APP_WEBID = hostname.includes("localhost")
+  ? "http://www.w3.org/ns/solid/terms#PublicOidcClient"
```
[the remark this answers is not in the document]
norbert989: @kay-kim this is the URI for the `PublicOidcClient` which was sent to me by Aaron, he also mentioned that the term needs to be added to the Solid vocabulary. 
I think it's best to ask him for more info on this. The usage here is simply to have it all working properly in dev localhost

pages/api/app.js
```diff
@@ -0,0 +1,34 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+export default function handler(req, res) {
+  const clientId = req.headers.host.includes("localhost")
+    ? "http://www.w3.org/ns/solid/terms#PublicOidcClient"
+    : `https://${req.headers.host}/api/app`;
+  res.status(200).json({
```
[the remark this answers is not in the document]
norbert989: I've addressed the top part of this @ajacksified
I think this could be extended later to handle different response types however, that could be another ticket?

components/login/index.jsx
```diff
@@ -48,6 +52,13 @@ export default function Login() {
   const INFO_TOOLTIP_TEXT = "This is where you signed up for a Solid Pod";
   const INFO_BUTTON_LABEL = "Where is your Pod hosted?";
 
+  const oidcSupported = useClientId(PROVIDER_IRI);
+
+  const authOptions = {
+    clientName: "Inrupt PodBrowser",
```
ajacksified, line 58: This should be a constant exported at the top of the file (most magic strings should be.)

components/login/provider/index.jsx
```diff
@@ -46,6 +50,10 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 const providers = getIdentityProviders();
 export const TESTCAFE_ID_LOGIN_FIELD = "login-field";
 const TESTCAFE_ID_GO_BUTTON = "go-button";
+const hostname = getCurrentHostname();
+const CLIENT_APP_WEBID = hostname?.includes("localhost")
```
ajacksified, line 54: `getCurrentHostname` should be refactored to return only the hostname and not the origin; and this should do an equality check, rather than "includes" (otherwise this would match `mylocalhost.com`, and other weird urls.)

ajacksified: This logic also seems to be re-used at least three times. Does a `isLocalhost` function make sense?

components/login/index.jsx
ajacksified: If `Inrupt PodBrowser` is used multiple times, it belongs in a constants file so it can be re-used.

pages/api/app.js
```diff
@@ -0,0 +1,39 @@
...
+    client_name: "Inrupt PodBrowser",
+    grant_types: ["refresh_token", "authorization_code"],
+    response_types: ["code"],
+  };
+}
+
+export default function handler(req, res) {
+  const clientId = req.headers.host.includes("localhost")
+    ? "http://www.w3.org/ns/solid/terms#PublicOidcClient"
+    : `https://${req.headers.host}/api/app`;
```
ajacksified, line 36: Same comment here re: localhost.

ajacksified (changes requested):

pages/api/app.js
norbert989: @ajacksified would you prefer this config to be removed and the defaults are taken?

components/login/index.jsx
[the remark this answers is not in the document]
kay-kim: Ah -- I see it's per the spec.
https://solid.github.io/authentication-panel/solid-oidc/#clientids-public-webid

components/login/index.jsx
norbert989: As per this comment, I moved `CLIENT_NAME` as well as `PUBLIC_OIDC_CLIENT` into a constants file as they are being reused.

norbert989: As per this comment, I moved `CLIENT_NAME` as well as `PUBLIC_OIDC_CLIENT` into a constants file as they are being reused.

components/login/provider/index.jsx
norbert989: added `getCurrentOrigin` as it was needed in order to build the `CLIENT_APP_WEBID` uri in the components, and edited `getCurrentHostname` to return just hostname. 

In stringHelpers added `isLocalhost(hostname)` which will return a `boolean` value based on an equality check.

pages/api/app.js
norbert989: `    grant_types: ["refresh_token", "authorization_code"],
    response_types: ["code"],`
    
    ☝️  I have removed the above two config lines and tested that everything works as expected.

ajacksified (approved):

PR #308, opened 2021-07-14
Refactor: confirmation dialog

VirginiaBalseiro:
In this PR:

- use confirmation dialog in the components where it's needed, so that it is appended in a way that can be tested

- write tests for interactions with the confirmation dialog

# To test: 

- Verify that the confirmation dialog works as intended in all the places where it's used:


ajacksified (approved):
Nice cleanup!

kevin-inrupt (approved):

PR #317, opened 2021-07-29
mocked checkOidc return value

norbert989:
This PR fixes code coverage after a previous merge into the main branch

### DIFF
--- a/components/login/provider/index.test.jsx
+++ b/components/login/provider/index.test.jsx
@@ -95,7 +95,7 @@ describe("ProviderLogin form", () => {
 
     const checkOidcSupportSpy = jest
       .spyOn(useClientId, "checkOidcSupport")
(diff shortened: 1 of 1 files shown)


components/login/provider/index.test.jsx
```diff
@@ -94,7 +94,9 @@ describe("ProviderLogin form", () => {
       label,
     });
 
-    const checkOidcSupportSpy = jest.spyOn(useClientId, "checkOidcSupport");
+    const checkOidcSupportSpy = jest
+      .spyOn(useClientId, "checkOidcSupport")
+      .mockImplementation(async () => true);
```
Vinnl, line 99: Doesn't really matter, but good to know that you can do this too :)

```suggestion
      .mockResolvedValue(true);
```

Vinnl (approved):

PR #327, opened 2021-08-17
Feature/consent request expiry date

norbert989:
# Consent request Expiry date input
- Added `dateInput` component that utilizes material ui pickers
- Added custom `forever` button that clears the date and sets the date input field value to a string `forever`

# Checklist


VirginiaBalseiro:
Can this purple header be removed? 

image

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+    setDateForever,
+  } = props;
+  const classes = useStyles();
+  const bem = useBem(classes);
+  const { consentRequest } = useContext(ConsentRequestContext);
+  const expirationDate = consentRequest?.expirationDate;
+  // FIXME: we will later fetch the expiry date from the consent details
+  const node = useRef();
+
+  const handleClick = (e) => {
```
VirginiaBalseiro, line 58: We have a hook for this `useOutsideClick`

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+    setDatepickerOpen(false);
+  };
+
+  useEffect(() => {
+    if (expirationDate) {
+      setSelectedDate(expirationDate);
+    }
+  }, [expirationDate, setSelectedDate]);
+
+  useEffect(() => {
```
VirginiaBalseiro, line 73: Can you explain why we're setting these here?

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+      document.removeEventListener("mousedown", handleClick);
+    };
+  });
+
+  return (
+    <div ref={node} className={bem("date-container")}>
+      <InputBase
+        classes={{ root: bem("date-input") }}
+        placeholder="Consent expiry date"
+        value={selectedDate ? format(selectedDate, "MMMM' 'd', 'Y") : "Forever"}
```
VirginiaBalseiro, line 87: If we're only using the date function here, is it possible for us to format the date manually instead of adding a library just for this?

components/consentRequestForm/styles.js
```diff
@@ -182,5 +201,48 @@ export default function styles(theme) {
...
+      borderRadius: 4,
+      boxShadow: `0 4px 8px 0 ${theme.palette.grey[500]}`,
+      overflow: "hidden",
+      display: "flex",
+      flexDirection: "column",
+      position: "absolute",
+      left: 0,
+      top: "calc(100% + 0.5rem)",
+      zIndex: 2,
+      backgroundColor: "white",
```
VirginiaBalseiro, line 240: minor thing, but we have white in the theme

VirginiaBalseiro (changes requested):

components/consentRequestForm/dateInput/index.jsx
norbert989: 👍  will update, wasn't aware the hook was in prism components

components/consentRequestForm/dateInput/index.jsx
norbert989: Setting the mouse click listener for click outside and cleaning it up as the component is removed. This will be replaced by the `useOutsideClick` hook

components/consentRequestForm/styles.js
norbert989: 👀  will update

components/consentRequestForm/dateInput/index.jsx
norbert989: we could of course. The format library is added as part of the dependencies for the `material-ui/pickers` so it wasn;t an extra addition however, it is up for us to decide if we just wanted to format the date manually

VirginiaBalseiro: Ah ok, I didn't know it was a dependency. How annoying.

norbert989:
> Can this purple header be removed?
> 
> image

Yes and no 😄 . We can remove it, of course, it will just mean that if a user wants to change the year they will have to use the arrows. Are we ok with that?

VirginiaBalseiro:
> > Can this purple header be removed?
> > image
> 
> Yes and no 😄 . We can remove it, of course, it will just mean that if a user wants to change the year they will have to use the arrows. Are we ok with that?

I think it’s not a major issue to use the arrows, I don’t expect people to be setting the expiration date for years after manually. But this is a question for @KyraAssaad

KyraAssaad:
> > > Can this purple header be removed?
> > > image
> > 
> > 
> > Yes and no 😄 . We can remove it, of course, it will just mean that if a user wants to change the year they will have to use the arrows. Are we ok with that?
> 
> I think it’s not a major issue to use the arrows, I don’t expect people to be setting the expiration date for years after manually. But this is a question for @KyraAssaad

I'm on the fence. I actually didn't even realize that you could click on the year in the purple header to change the year (bad UX). The only situation I could imagine is if an app requested access for, say, two years from now, and you wanted to change it to one month from now instead. I hypothesize that not a lot of people will be using this anyway, but just in case, let's keep it for now (but I reserve the right to change my mind later 😜 )

KyraAssaad:
Inrupt_PodBrowser

@norbert989 could you add a couple extra px of padding so the numbers don't get cutoff on months that have 6 rows please

KyraAssaad (changes requested):
Inrupt_PodBrowser

@norbert989 could you add a couple extra px of padding so the numbers don't get cutoff on months that have 6 rows please

components/consentRequestForm/styles.js
norbert989: @VirginiaBalseiro the theme does not seem to have `white` configured in the palette. Checking in prism even there `color: white` is used instead. Can we keep it as is in this PR and raise a ticket to change that? It would require updating both prism and PB usages as there are multiple in both repos 😬

VirginiaBalseiro: It's actually `theme.palette.background.paper`

KyraAssaad (approved):
Yay!

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+    setSelectedDate,
+    datepickerOpen,
+    setDatepickerOpen,
+    handleDateChange,
+    setDateForever,
+  } = props;
+  const classes = useStyles();
+  const bem = useBem(classes);
+  const { consentRequest } = useContext(ConsentRequestContext);
+  const expirationDate = consentRequest?.expirationDate;
```
VirginiaBalseiro, line 54: This is still relying on destructuring JSON-LD. I suggest adding a function to the model `getExpiryDate` that for now, just does the same that here and returns the value, but later it will be easier to swap for `solid-client` functions.

components/consentRequestForm/dateInput/index.test.jsx
```diff
@@ -0,0 +1,87 @@
...
+import { MuiPickersUtilsProvider } from "@material-ui/pickers";
+import { renderWithTheme } from "../../../__testUtils/withTheme";
+import { ConsentRequestProvider } from "../../../src/contexts/consentRequestContext";
+import ConsentRequestFrom from "../index";
+import {
+  TESTCAFE_ID_FOREVER_BUTTON,
+  TESTCAFE_ID_DATE_PICKER_CALENDAR_BUTTON,
+} from "./index";
+
+describe("DateInput component", () => {
```
VirginiaBalseiro, line 34: Is this different somehow from the request we get from `getRequestDetails`?

components/consentRequestForm/styles.js
VirginiaBalseiro: It's a minor thing but helpful if we ever want to change the theme. We should not be using hard coded colors for this reason (unless they're not meant to change with the theme, but background and font colors usually do).

package.json
```diff
@@ -26,17 +26,20 @@
...
     "@inrupt/solid-client": "^1.7.0",
     "@inrupt/solid-client-authn-browser": "^1.8.2",
     "@inrupt/solid-ui-react": "^2.3.1",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
+    "@material-ui/pickers": "^3.3.10",
     "@sentry/node": "^6.1.0",
     "@sentry/react": "^6.1.0",
     "@solid/lit-prism-patterns": "^0.13.5",
```
VirginiaBalseiro, line 40: Let's keep all prism packages in the same version (latest)

VirginiaBalseiro (changes requested):

components/consentRequestForm/dateInput/index.test.jsx
norbert989: oh that's a miss from my side. It was just a simplified one with one single section as I wrote it. Was meant to update it to the mock one. I'll make this change

VirginiaBalseiro (approved):

PR #340, opened 2021-09-01
added enw page layout for agent resource access

norbert989:
# Applied new layout for the Agent Access resource page

# Checklist


components/profile/personAvatar/styles.js
```diff
@@ -0,0 +1,45 @@
...
+const styles = (theme) => {
+  return createStyles(theme, ["back-to-nav", "input"], {
+    avatar: {
+      width: "64px",
+      height: "64px",
+    },
+    avatarText: {
+      fontFamily: theme.typography.h1.fontFamily,
+      fontSize: "1.5rem",
+      color: "#000",
```
VirginiaBalseiro, line 33: Can we pull this from the theme?

VirginiaBalseiro (approved):
looks good!

VirginiaBalseiro (changes requested):
I just noticed this doesn't work with apps, you can see how the profile link is built here: https://github.com/inrupt/pod-browser/pull/313

VirginiaBalseiro:
Also, I just realized we don't have a specific subtask for the styling of the resource table so it should be done as part of this.

VirginiaBalseiro (changes requested):
When I visit a person's profile I get directed to the `app` path and therefore the app profile is shown instead of the person's (the path should be `person`):

image

VirginiaBalseiro (changes requested):
Also missing the message "Agent has access to these resources" above the table: 

image

And could we add a spinner while resources are loading in the table? It's a bit disconcerting when one first loads the page and it takes a bit to display the resources.

components/profile/appAvatar/styles.js
```diff
@@ -0,0 +1,54 @@
...
+      height: "64px",
+      background: "transparent",
+    },
+    appAvatar: {
+      textAlign: "center",
+      padding: theme.spacing(1),
+      borderRadius: "50%",
+      fontSize: theme.typography.h5.fontSize,
+      color: theme.palette.primary.main,
+      backgroundColor: "#F1EDFF",
```
VirginiaBalseiro, line 37: Can we put this color in a constant inside the same file? This way we can track later which colors we're reusing and should be added to the theme.

VirginiaBalseiro (changes requested):

components/profile/appAvatar/styles.js
norbert989: This color was actually used a few times before so I added the const in all places and marked with a TODO

VirginiaBalseiro (approved):

KyraAssaad (dismissed):
It looks great! A few things: 
1. It looks like the table sorting on the Access column isn't working correctly
Inrupt_PodBrowser
I'd be happy to remove the sorting on the column if that's easier.

2. There's a weird thing that happens when you are scrolled to the bottom of the page and then click on a resource to open the sidebar- once the content in the sidebar has loaded, the top of the sidebar where the Close button is disappears. You can see it again if you scroll up a bit, but it's disorienting if you don't know to do that. I can't replicate the behaviour on the Files page- the top of the sidebar with the Close button is always visible no matter how far you scroll down the page (although it seems that when you open the sidebar, the page jumps back up to the top so that might be why). Let me know if you want to chat quickly so I can show you. If it's not an easy fix, we can create a ticket to address it later.

PR #346, opened 2021-09-13
Feature/agent access resource path

norbert989:
# Added resource path for nested folders/files

- Added resource path to table rows
- Added resource path to details drawer

# Checklist

### DIFF
--- a/components/pages/privacy/resourceAccess/resourceAccessDrawer/__snapshots__/index.test.jsx.snap
+++ b/components/pages/privacy/resourceAccess/resourceAccessDrawer/__snapshots__/index.test.jsx.snap
@@ -36,9 +36,16 @@ exports[`ResourceDrawer Renders the ResourceDrawer with correct title and access
             <span
               class="PodBrowser-icon-folder PodBrowser-access-details--icon"
             />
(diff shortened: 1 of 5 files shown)


KyraAssaad (changes requested):
Ahah...ha...ha.... I broke it 😁 

Inrupt_PodBrowser

Changes:
1. Let's make the Name column have a max-width of 650px. 
2. In the case that it overflows, we should prioritize displaying the name of the resource, and truncate the rest of the path with ellipses. For example, it would look like "...[name of resource]". In the event that even the name of the resource is too long to fit 650px, then the front of the name should be truncated and the end of the name should be displayed
3. In the sidebar, let's wrap the name of the path the same way we wrap the name of the resource
4. Is it possible to remove the HTML spaces conversion from the path, so we don't get the %20 cluttering up the names?
5. Can we also please top align the resource icon with the name of the resource
Inrupt_PodBrowser

VirginiaBalseiro (approved):

KyraAssaad (approved):
You did it! Nicely done.

kevin-inrupt (approved):

PR #374, opened 2021-11-04
Support both legacy and latestACRs

NSeydoux:
The behavior of ESS changed between versions 1.1 and 1.2. In particular, it is no longer possible to reference external policies in ACRs. This PR implements the changes needed to support both from the same codebase.

Feature detection is implemented based on configuration detection as specified in the latest ACP spec, with a small caveat: we use an OPTIONS request instead of a HEAD request. A hook is now exposed to call this feature detection function, and based on its response value the corresponding AccessControl strategy is initialized with a `isLegacy` value. This value is used to switch between the current (legacy) behaviour that maintains backwards-compatibility with ESS 1.1, currrently deployed in pod-spaces, and the latest behaviour implemented on ESS 1.2. Most of the code in this PR comes from either this switch and testing the compatibility with both behaviours.

# Checklist

- [X] All acceptance criteria are met.
- [X] Relevant documentation, if any, has been written/updated.


src/hooks/useAccessControl/index.test.jsx
```diff
@@ -51,6 +53,8 @@ describe("useAccessControl", () => {
       .spyOn(accessControlFns, "getAccessControl")
       .mockResolvedValue(accessControl);
     mockedPoliciesContainerUrlHook.mockReturnValue(null);
+    const mockedLegacyAcpHook = jest.requireMock("../useIsLegacyAcp");
```
VirginiaBalseiro, line 56: You can just import the hook as normal at the top of the file and then do `const mockedLegacyAcpHook = useIsLegacyAcp`. Then when you mock the return value you don't need to add `default` to it.

src/hooks/usePoliciesContainerUrl/index.test.js
```diff
@@ -21,28 +21,38 @@
...
+      const { result } = renderHook(() => usePoliciesContainerUrl(null));
+      expect(result.current).toBeNull();
+    });
 
-  it("returns response when podRootUrl is finished", async () => {
-    mockedPodRootUri.mockReturnValue(podRootUrl);
-    const { result } = renderHook(() => usePoliciesContainerUrl());
-    expect(result.current).toBe(policiesContainerUrl);
+    it("returns response when podRootUrl is finished", async () => {
+      const clientModule = jest.requireMock("@inrupt/solid-client");
```
VirginiaBalseiro, line 47: Same as in the other test file.

VirginiaBalseiro (changes requested):
The code looks good, I just left a few comments that can be ignored.

When testing though, I find the following: 
On 1.1: 

- With an existing container, the existing permissions are correctly displayed. I can't, however, make any changes to the policies. Adding or removing an agent doesn't seem to error out, but the agents are unchanged in the panel, even when refreshing. I see a `PATCH` with the expected content (and agents) request to the ACR that seems to succeed. However in subsequent `GET`s of the ACR, the agents are not in the policy. I do see some `401`s when getting the resource, but then the same requests appear to succeed. 
- With a new container, same behavior as above. I cannot change the policy. 

On 1.2: 
- It does seem to add the agent and I briefly see it has been added to the list, but then I get a `401` on the container (which is in my pod) and the page goes blank. Then anytime I try to see permissions for that resource, I get a 404 when requesting: `https://id.dev-next.inrupt.com/.well-known/solid` and the page goes blank.

NSeydoux:
Regarding the issues you reported: 
- On 1.1, you are absolutely correct, I'm not sure how I missed this in the first place, 50b993ab475548be05271f361120542c3b6b4968 should resolve the problem.
- On 1.2, the 401 should be expected I think (it's part of the authorization server discovery to issue an unauthenticated request. However, on my local deployment I can see the added agents, will test on Vercel. And b0fa77b2161cf83b4a219d18139afb9de1521703 should resolve the second issue you describe: `.well-known/solid` discovery cannot be done from the WebID, so I added a storage discovery step instead.

src/hooks/useAccessControl/index.test.jsx
NSeydoux: 30bbd48

src/hooks/usePoliciesContainerUrl/index.test.js
NSeydoux: 30bbd48

VirginiaBalseiro (approved):

PR #382, opened 2021-11-05
Fix/nss no control

VirginiaBalseiro:
This PR fixes bug #SOLIDOS-1062.

`isAcpControlled` is async so we needed to await it before determining if a resource is Wac controlled or not.

To test: 

- Log in with NSS
- Verify the "No control" yellow warning is not displayed and the pod contents are accessible.


components/resourceDetails/index.jsx
```diff
@@ -93,8 +95,15 @@ export default function ResourceDetails({
...
-  const useWac = isWac(datasetUrl, dataset, fetch);
+
+  useEffect(() => {
+    (async () => {
+      const acp = await isAcp(datasetUrl, fetch);
+      const wac = await isWac(datasetUrl, dataset, fetch);
+      setUseAcp(acp);
+      setUseWac(wac);
+    })();
+  }, [datasetUrl, dataset, fetch]);
```
NSeydoux, line 106: What is the difference between doing this and using the `useSWR` hook, and is one approach better than the other in this context ?

VirginiaBalseiro: We could have separate hooks (`useAcp`, `useWac`) for each value.

src/accessControl/index.js
```diff
@@ -36,15 +36,19 @@ export function hasAccess(resourceInfo) {
...
 }
 
-export function isAcp(resourceUrl, fetch) {
-  return resourceUrl && acp3.isAcpControlled(resourceUrl, { fetch });
+export async function isAcp(resourceUrl, fetch) {
+  const isAcpControlledResource = await acp3.isAcpControlled(resourceUrl, {
+    fetch,
+  });
+  return resourceUrl && isAcpControlledResource;
 }
```
NSeydoux, line 44:
```suggestion
export async function isAcp(resourceUrl, fetch) {
  return resourceUrl && await acp3.isAcpControlled(resourceUrl, {
    fetch,
  });
}
```
What about this, without the intermediary variable ?

src/accessControl/index.js
```diff
@@ -36,15 +36,19 @@ export function hasAccess(resourceInfo) {
...
-export function isWac(resourceUrl, resourceInfo, fetch) {
+export async function isWac(resourceUrl, resourceInfo, fetch) {
+  const isAcpControlledResource = await acp3.isAcpControlled(resourceUrl, {
+    fetch,
+  });
   return (
-    resourceUrl &&
-    !acp3.isAcpControlled(resourceUrl, { fetch }) &&
-    hasAccessibleAcl(resourceInfo)
+    resourceUrl && !isAcpControlledResource && hasAccessibleAcl(resourceInfo)
```
NSeydoux, line 51: Ditto

NSeydoux (approved):

PR #397, opened 2021-11-25
Fix/delete resource policy resource

VirginiaBalseiro:
This PR fixes bug #PB-1109.

- we were not correctly checking that we were using legacy ACP, therefore the policy URL was not correctly found and the policy resource was not being deleted

- updating several hooks to get the correct policies containers and URL

- update tests

### DIFF
--- a/src/accessControl/acp/index.js
+++ b/src/accessControl/acp/index.js
@@ -32,7 +32,6 @@ import {
   saveSolidDatasetAt,
   setThing,
 } from "@inrupt/solid-client";
-import LinkHeader from "http-link-header";
 import {
   ACL,
   createAccessMap,
@@ -40,7 +39,6 @@ import {
   isEmptyAccess,
 } from "../../solidClientHelpers/permissions";
 import { chain, createResponder } from "../../solidClientHelpers/utils";
-// eslint-disable-next-line import/no-cycle
 import { getOrCreateDatasetOld } from "../../solidClientHelpers/resource";
 import {
   isCustomPolicy,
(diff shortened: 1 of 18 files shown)


solid-akb (approved):
Awesome!

src/accessControl/acp/index.js
```diff
@@ -40,14 +40,13 @@ import {
   isEmptyAccess,
 } from "../../solidClientHelpers/permissions";
 import { chain, createResponder } from "../../solidClientHelpers/utils";
-import {
-  deletePoliciesContainer,
-  getOrCreateDatasetOld,
-} from "../../solidClientHelpers/resource";
+// eslint-disable-next-line import/no-cycle
```
ajacksified, line 43: We should definitely fix this (separately.)

src/hooks/useAccessControl/index.jsx
```diff
@@ -20,13 +20,15 @@
...
 import { useSession } from "@inrupt/solid-ui-react";
+import { getSourceUrl } from "@inrupt/solid-client";
 import useSWR from "swr";
 import { getAccessControl } from "../../accessControl";
 import usePoliciesContainerUrl from "../usePoliciesContainerUrl";
 import useIsLegacyAcp from "../useIsLegacyAcp";
 
 export default function useAccessControl(resourceInfo, swrOptions = {}) {
-  const policiesContainerUrl = usePoliciesContainerUrl(resourceInfo);
+  const resourceUrl = resourceInfo ? getSourceUrl(resourceInfo) : null;
```
ajacksified, line 30: `getSourceUrl` returns string | null, so you shouldn't need this ternary.

src/hooks/useIsLegacyAcp/index.js
```diff
@@ -28,7 +28,7 @@ import { hasAcpConfiguration } from "../../accessControl/acp/index";
 export default function useIsLegacyAcp(resourceInfo) {
   const { fetch } = useSession();
   return useSWR(
-    resourceInfo ? acp.getLinkedAcrUrl(resourceInfo) : null,
+    [resourceInfo ? acp.getLinkedAcrUrl(resourceInfo) : null, fetch],
```
ajacksified, line 31: ...although this one returns undefined, instead of null, so this ternary belongs here. I've started a conversation in the SDK team to align on when we return null vs undefined.

src/models/policy/index.js
```diff
@@ -97,3 +98,17 @@ export function isCustomPolicy(type) {
...
 }
+
+export const deletePoliciesContainer = async (containerIri, fetch) => {
+  try {
+    await deleteContainer(containerIri, { fetch });
+  } catch (err) {
+    if (
+      !isHTTPError(err.message, 409) &&
+      !isHTTPError(err.message, 404) &&
+      !isHTTPError(err.message, 403)
```
ajacksified, line 109: Do we only want to specifically catch these three, or should this catch any error responses?

src/accessControl/acp/index.js
VirginiaBalseiro: I moved the function causing the cycle to a `helpers` subfolder, let me know if you have a better idea for this.

src/models/policy/index.js
VirginiaBalseiro: We want to ignore these 3 and throw and error if only if it isn't one of these.

kevin-inrupt (approved):

ajacksified (approved):

PR #404, opened 2022-01-03
Fix/edit 1 2 profiles

VirginiaBalseiro:
This PR fixes bug #PB-1041.

In this PR, we switch to using the profile from `SessionContext` so that we can determine whether it can be edited and display the appropriate profile document for 1.1 and 1.2. In the case of 1/2 for now we are only getting the first profile if there are multiple available. 


solid-akb (approved):
This looks like a tricky PR - nice work!

kevin-inrupt (approved):

PR #405, opened 2022-01-03
Copy pod location

solid-akb:
# New feature description
[PB-1075](https://inrupt.atlassian.net/browse/PB-1075)

https://user-images.githubusercontent.com/94387648/155302127-7c9a1ac8-6d83-48d0-b16e-815e66f569aa.mp4


VirginiaBalseiro (changes requested):
Unfortunately, when I try to click on the pod indicator to display the dropdown, the page blanks. This is because the Tooltip component needs one child only to forward the ref, so you could try wrapping all the children inside the Tooltip with a `span`. 

Also: 
- Can we add some tests (i.e.: that the button calls the copy function, that the tooltip is displayed as expected, etc)?
- There seems to be an issue with the package.json and package-lock.json being out of sync which is causing the build to fail.

solid-akb:
Thanks for the feedback!

I will fix the tooltip thing. 

For testing, I ran into an issue where I couldn’t “click” on the list item when the page loaded because when the page loads the drop down list doesn’t render (it renders when you click on the pod uri). Do you have any suggestions on how to work around that?

Package JSON lock file I think/hope will get fixed if I run Npm ci before committing. I wasn’t doing that before!

Sent from my iPhone

> On Jan 5, 2022, at 06:12, Virginia Balseiro ***@***.***> wrote:
> 
> ﻿
> @VirginiaBalseiro requested changes on this pull request.
> 
> Unfortunately, when I try try click on the pod indicator to display the dropdown, the page blanks. This is because the Tooltip component needs one child only to forward the ref, so you could try wrapping all the children inside the Tooltip with a span.
> 
> Also:
> 
> Can we add some tests (i.e.: that the button calls the copy function, that the tooltip is displayed as expected, etc)?
> There seems to be an issue with the package.json and package-lock.json being out of sync which is causing the build to fail.
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> You are receiving this because you authored the thread.

-- 
This e-mail, and any attachments thereto, is intended only for use by the 
addressee(s) named herein and may contain legally privileged, confidential 
and/or proprietary information. If you are not the intended recipient of 
this e-mail (or the person responsible for delivering this document to the 
intended recipient), please do not disseminate, distribute, print or copy 
this e-mail, or any attachment thereto. If you have received this e-mail in 
error, please respond to the individual sending the message, and 
permanently delete the email.

VirginiaBalseiro:
> For testing, I ran into an issue where I couldn’t “click” on the list item when the page loaded because when the page loads the drop down list doesn’t render (it renders when you click on the pod uri). Do you have any suggestions on how to work around that?

You will have to simulate a click, then await the list to be rendered, and take it from there. We can pair on this if you'd like and/or I can provide examples of where we do similar things.

solid-akb:
That makes sense. Ok. I’ll try and give it a first pass. I’m getting a Covid test so I’m not sure what my timing will be like this morning but if I’m free and you’re free it would be fun to do a PP.

Sent from my iPhone

> On Jan 5, 2022, at 07:17, Virginia Balseiro ***@***.***> wrote:
> 
> ﻿
> For testing, I ran into an issue where I couldn’t “click” on the list item when the page loaded because when the page loads the drop down list doesn’t render (it renders when you click on the pod uri). Do you have any suggestions on how to work around that?
> 
> You will have to simulate a click, then await the list to be rendered, and take it from there. We can pair on this if you'd like and/or I can provide examples of where we do similar things.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> You are receiving this because you authored the thread.

-- 
This e-mail, and any attachments thereto, is intended only for use by the 
addressee(s) named herein and may contain legally privileged, confidential 
and/or proprietary information. If you are not the intended recipient of 
this e-mail (or the person responsible for delivering this document to the 
intended recipient), please do not disseminate, distribute, print or copy 
this e-mail, or any attachment thereto. If you have received this e-mail in 
error, please respond to the individual sending the message, and 
permanently delete the email.

components/podIndicator/index.test.jsx
```diff
@@ -69,3 +78,30 @@ describe("closeHandler", () => {
     expect(setAnchorEl).toHaveBeenCalledWith(null);
   });
 });
+
+describe("copy pod uri", () => {
+  jest.setTimeout(3000);
```
VirginiaBalseiro, line 83: This shouldn't be necessary.

solid-akb: agreed - you can take it out. I was trying a few things to see why tests might be passing locally but not here.

VirginiaBalseiro (approved):

components/podIndicator/index.jsx
```diff
@@ -168,6 +178,32 @@ export default function PodIndicator() {
...
+            >
+              <ListItem
+                button
+                key="copy-text"
+                onClick={handlePodCopyClick}
+                classes={{ root: bem("menuItem") }}
+                data-testid={TESTCAFE_ID_POD_INDICATOR_COPY}
+              >
+                <ListItemIcon classes={{ root: bem("itemIcon") }}>
+                  <Icons
```
KyraAssaad, line 199: The icon is not displaying. It looks like in line 174 for the **Change Pod** item it uses `<i>` but here it's using `<Icons>`

Inrupt_PodBrowser

KyraAssaad (changes requested):
The menu isn't closing after the link is clicked, as proposed by Kevin in https://inrupt.atlassian.net/browse/APPS-1075?focusedCommentId=20839

components/podIndicator/index.jsx
solid-akb: Missed Kevin's comment but have added that feature! The icon is a new icon added to Prism and is dependent on us being able to use a later version of Prism which is what we are working on in APPS-1145. <i> tags are how we pull the icons from Material UI, which our version doesn't contain the "content-copy" icon that we want to use here. <Icons> tags are when we get the icon from Prism (example in line 112). Want to keep this PR open until PRISM gets updated (which has to be done anyways) instead of spending time trying to figure out why "content-copy" is not available to us in Material UI.

KyraAssaad (approved):
Beauty!

PR #407, opened 2022-01-13
Use client identifiers with abritrary provider

NSeydoux:
This enforces that the client identifier for PodBrowser is used with any OIDC provider supporting it, and not only with the default one. Previously, providing a custom Solid-OIDC provider would default to dynamic client registration.


NSeydoux:
This doesn't fix what I think it did, so I'll create a bug ticket to investigate properly.

PR #410, opened 2022-01-21
Fix: display confirmation dialog when clicking Delete contact button

VirginiaBalseiro:
This PR fixes bug #APPS-1107.

The ConfirmationDialog was not added to the Drawer component and therefore not displayed. We previously attempted to display in in the App component, but this led to issues with testing, although we might want to revisit that idea in the future to avoid having to add the component to each component where it is used (and having these type of bugs).


ajacksified (approved):
> although we might want to revisit that idea in the future to avoid having to add the component to each component where it is used

Definitely, would you create a ticket for refactoring the confirmation dialog?

kevin-inrupt (approved):

PR #412, opened 2022-01-26
Remove public client; update solid-client-authn

ajacksified:
This fixes three issues:

* Removes the use of OIDCPublicClient from localhost, which is being deprecated. Instead, localhost now uses dynamic client registration.
* Upgrades solid-client-authn, which now includes the webid OIDC scope.
* Fixes an issue in localhost developing against the latest dev version of ESS, causing an infinite loop due to an OIDC error.


solid-akb (approved):
Looks good!

PR #414, opened 2022-01-28
Chore/fix vulns

VirginiaBalseiro:
- Fix vulnerabilities via `npm audit fix`
- Fix additional vulnerabilities by upgrading `solid-client-access-grants` to latest version
- Replace deprecated methods from `solid-client-access-grants`

PR #416, opened 2022-02-01
bumped up solid-client & solid-client-access-grants

solid-akb:
Running into Consent flow and it was recommended we update our libraries. Didn't fix the issue but is still good practice to update the libraries. With a quick run through, no flows break and tests are good.


solid-akb:
Covered in https://github.com/inrupt/pod-browser/pull/414

PR #420, opened 2022-02-03
Fix: remove previousPage from localStorage after use

VirginiaBalseiro:
This PR fixes bug #APPS-1172.

We now remove `previousPage` from localStorage after using it to redirect the user back to the previous page after logging in.


ajacksified:
@VirginiaBalseiro #419 does this, but adds tests and does a bit of cleanup too.

PR #424, opened 2022-02-07
Consent styling

solid-akb:
This PR fixes bug #1090.

Updated styling to help with overflowing content. Essentially the URIs are long strings, and the div looks for a white space to break it up. If there is no white space it run outside of the div. The word-break css attribute allows the string be broken into lines if necessary without the string containing whitespace.


solid-akb:
Issues pulling in other irrelevant files. New PR is https://github.com/inrupt/pod-browser/pull/425/files

PR #425, opened 2022-02-08
Consent form text overflow

solid-akb:
This PR fixes bug #1090.

Updated styling to help with overflowing content. Essentially the URIs are long strings, and the div looks for a white space to break it up. If there is no white space it run outside of the div. The word-break css attribute allows the string be broken into lines if necessary without the string containing whitespace.

### DIFF
--- a/components/consentRequestForm/styles.js
+++ b/components/consentRequestForm/styles.js
@@ -202,11 +202,17 @@ export default function styles(theme) {
     "agent-name": {
       display: "flex",
       justifyContent: "center",
+      textOverflow: "ellipsis",
+      overflow: "hidden",
+      wordBreak: "break-word",
     },
     purpose: {
       display: "flex",
       justifyContent: "flex-start",
       padding: "1rem",
+      textOverflow: "ellipsis",
+      overflow: "hidden",
+      wordBreak: "break-word",
     },
 
     "date-container": {
@@ -307,6 +313,9 @@ export default function styles(theme) {
     },
     "purposes-container": {
       padding: "2rem 0",
+      textOverflow: "ellipsis",
+      overflow: "hidden",
+      wordBreak: "break-word",
     },
   };
 }
--- a/components/consentRequestForm/styles.js
+++ b/components/consentRequestForm/styles.js
@@ -202,19 +202,13 @@ export default function styles(theme) {
     "agent-name": {
       display: "flex",
       justifyContent: "center",
-      textOverflow: "ellipsis",
-      overflow: "hidden",
       wordBreak: "break-word",
-      whiteSpace: "nowrap"
     },
     purpose: {
       display: "flex",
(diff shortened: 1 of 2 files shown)


components/consentRequestForm/styles.js
```diff
@@ -202,11 +202,17 @@ export default function styles(theme) {
     "agent-name": {
       display: "flex",
       justifyContent: "center",
+      textOverflow: "ellipsis",
```
VirginiaBalseiro, line 205: In order for ellipsis to show up, we need to have also `white-space` set to `nowrap`.

components/consentRequestForm/styles.js
```diff
@@ -202,11 +202,17 @@ export default function styles(theme) {
...
+      overflow: "hidden",
+      wordBreak: "break-word",
     },
     purpose: {
       display: "flex",
       justifyContent: "flex-start",
       padding: "1rem",
+      textOverflow: "ellipsis",
```
VirginiaBalseiro, line 213: In order for ellipsis to show up, we need to have also `white-space` set to `nowrap`.

components/consentRequestForm/styles.js
```diff
@@ -307,6 +313,9 @@ export default function styles(theme) {
     },
     "purposes-container": {
       padding: "2rem 0",
+      textOverflow: "ellipsis",
```
VirginiaBalseiro, line 316: In order for ellipsis to show up, we need to have also `white-space` set to `nowrap`.

VirginiaBalseiro (changes requested):
At 320px wide, the Select link is still overflowing. Do we want to somehow wrap that too, maybe switch to column when the button starts to overflow? 

image

Not sure how common but there are a number of devices which fall under this width: 

https://viewportsizes.mattstow.com/?filter=

cc @KyraAssaad

solid-akb:
@VirginiaBalseiro I put in the ellipses for good measure as a fall-back, but it's not actually what we want so I removed the overflow attributes. My local testing was working because the ellipses weren't actually kicking in without the no-wrap. This user story is to wrap the text so that it doesn't run off screen. I believe there are hardening stories for if the URI is so long and how to handle it (ie. at how many lines do we stop wrapping the text and start just adding the ellipses)

VirginiaBalseiro:
> @VirginiaBalseiro I put in the ellipses for good measure as a fall-back, but it's not actually what we want so I removed the overflow attributes. My local testing was working because the ellipses weren't actually kicking in without the no-wrap. This user story is to wrap the text so that it doesn't run off screen. I believe there are hardening stories for if the URI is so long and how to handle it (ie. at how many lines do we stop wrapping the text and start just adding the ellipses)

In that case let's remove the ellipsis altogether since without `overflow` set to hidden, this will be ignored.

PR #435, opened 2022-02-23
Switch localhost to using a hosted clientid

ajacksified:
This removes DCR from localhost and uses a statically hosted clientid file for authentication.


VirginiaBalseiro (approved):

solid-akb (approved):
Sweet!

PR #430, opened 2022-02-17
Apps 1123 account dropdown title fix

daytonn:
## Description
This PR fixes the recently changed profile which breaks the account dropdown UI element ([1123](https://inrupt.atlassian.net/browse/APPS-1123)).

## Changes
 - Change the session test helper to have configurable profile documents
 - Some minor cleanup in `container/index.jsx` just because I was in that file and couldn't help myself
 - Also some minor cleanup in `usePodIrisFromWebId`
 - Changed `fetchProfile` to accommodate the new profile workflow

## Testing
 - Log into Pod Browser via another provider
 - Use `https://openid.dev-next.inrupt.com`
 - Verify you see your account name or nickname in the account menu
 
## Commit checklist

- [x] All acceptance criteria are met.
- [x] Includes tests to ensure functionality and prevent regressions.
- [ ] Relevant documentation, if any, has been written/updated.
- [ ] The changelog has been updated, if applicable.

## Notes
I couldn't resist and made some minor cleanup to a couple files here. I realize that some may find this less than ideal so I'll gladly remove that code if preferred. Otherwise, I'd like to just sneak in some minor clean up.

## Screenshots/captures
profile-menu

PR #436, opened 2022-02-23
Conatiner component cleanup

daytonn:
## Description
This PR is just some minor cleanup I couldn't resist. It was pulled from another PR. I removed some of the logic from the JSX section of the component to be pulled up in a place where the logic is easier to read and reason about.

## Changes
- Added a predicate function to clear up the logic regarding a missing iri
- Added a function that can conditionally render the spinner or the content of the container
- Added a function to contain the logic of possibly rendering a warning message

## Testing
 - Pure refactor, there should not be any noticeable changes.

## Interested parties
@VirginiaBalseiro 

## Notes
This was pulled from [another PR](https://github.com/inrupt/pod-browser/pull/430) by @VirginiaBalseiro's request. This also represents a bit of a style preference. I tend to prefer all logic be done outside of the JSX block in the return body. This helps keep the JSX readable and straight-forward and promotes testability of the extracted code.

### DIFF
--- a/components/container/index.jsx
+++ b/components/container/index.jsx
@@ -49,6 +49,32 @@ import { isHTTPError } from "../../src/error";
 import { locationIsConnectedToProfile } from "../../src/solidClientHelpers/profile";
 import { isContainerIri } from "../../src/solidClientHelpers/utils";
 
+function isNotAContainerResource(iri, container) {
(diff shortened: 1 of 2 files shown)


solid-akb (approved):
Thanks for doing cleanup!

PR #437, opened 2022-02-25
Improve Session Handling: Block rendering until session state is known

ThisIsMissEm:
This is a bit of a hack of something that should really be done in solid-ui-react: Currently the SessionProvider will render it's children when the session is in an unknown state, this leads to all sorts of UI glitches and double renders and such.

In this PR, I've added a child to the SessionProvider (in lieu of modifying solid-ui-react's SessionProvider), such that if the session is currently loading, then we show a full-page spinner. 

This also attempts to improve the error handling of sessions, such that we correctly catch errors and handle them in some "meaningful" way (currently I'm just rendering a div, but the idea is to render some sort of 500 "something went wrong, please reload" with a crash report sent. By handling errors from the Session, we may also catch the infinite redirects error.

If I have time, I'll try to write up something for the @inrupt/sdk-owners to implement this directly in `SessionProvider` by accepting a `loading` and `error` component prop, as well as accepting additional callbacks (onLogin, onLogout, etc), but this is a proof of concept that works and should make the application more reliable / responsive.

For the pod-browser team, this means you no longer need to check if you're logging in or not: you can just grab the session. You may want to write a `<AuthenticatedPage>` component that checks for the session and renders something other than the requested page if you're not authenticated.

Here's what it looks like:

https://user-images.githubusercontent.com/30827/155797050-f95539ca-1505-4339-8ff9-6e1d76ba6ba2.mov

---

This PR fixes bug #.

# New feature description

# Checklist


ThisIsMissEm:
When testing this, I did get the nextjs dev server segfaulting quite a lot on node v14, but not on v16 — I'm not sure what the issue is exactly. Maybe I introduced a regression, I'm not sure.

pages/_app.jsx
```diff
@@ -92,11 +94,11 @@ export function hasSolidAuthClientHash() {
   return false;
 }
 
-const updateHistory = (prevState, path) => {
+function updateHistory(prevState, path) {
```
ThisIsMissEm, line 97: @VirginiaBalseiro eslint was complaining about this

pages/_app.jsx
```diff
@@ -144,28 +146,24 @@ export default function App(props) {
       <MatomoProvider value={matomoInstance}>
         <StylesProvider jss={jss}>
           <ThemeProvider theme={theme}>
-            <SessionProvider
-              sessionId="pod-browser"
-              restorePreviousSession
-              onSessionRestore={async (url) => {
```
ThisIsMissEm, line 150: This is handled in `SessionReady` now, along with handlers for logout and errors.

src/SessionReady/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+import Spinner from "../../components/spinner";
+
+export default function SessionReady({ children }) {
+  const [sessionError, setSessionError] = useState();
+  const router = useRouter();
+  const session = useSession();
+
+  useEffect(() => {
+    const instance = getDefaultSession();
```
ThisIsMissEm, line 37: `useSession` / `SessionProvider` uses `getDefaultSession` under the hood, which is a singleton, this means when we mount SessionReady we can add those listeners.

src/SessionReady/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+import { getDefaultSession } from "@inrupt/solid-client-authn-browser";
+
+import Spinner from "../../components/spinner";
+
+export default function SessionReady({ children }) {
+  const [sessionError, setSessionError] = useState();
+  const router = useRouter();
+  const session = useSession();
+
+  useEffect(() => {
```
ThisIsMissEm, line 36: This `useEffect` should probably have [a cleanup](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1) such that we do `instance.off` on each of "error", "onSessionRestore" and "logout" when the `SessionReady` component is unmounted (even though I don't think it should be unmounted, though I did get a warning about too many eventemitter listeners in testing, so maybe it's related

src/SessionReady/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+      console.log("onSessionRestore");
+      try {
+        await router.push(url);
+      } catch (err) {
+        console.error(err);
+      }
+    });
+
+    instance.onLogout(async () => {
+      console.log("onLogout");
```
ThisIsMissEm, line 53: This needs to go, it's more debug stuff to make sure things are firing correctly.

pages/_app.jsx
```diff
@@ -144,28 +146,24 @@ export default function App(props) {
...
-
-                      <main className={bem("app-layout__main")}>
-                        <Component {...pageProps} history={history} />
-                      </main>
-                    </div>
-                    <Notification />
-                  </ConfirmationDialogProvider>
-                </AlertProvider>
-              </FeatureProvider>
+            <CssBaseline />
```
ThisIsMissEm, line 149: As far as I know this isn't Session dependent. I'd encourage reordering the providers here based on which updates most frequently as being the deepest provider.

ThisIsMissEm:
If you happen to notice a "flash of serif" on the login page, then it looks like it's down to how we're loading our webfonts, though I'm not sure how I managed to expose this as an issue.

daytonn:
> When testing this, I did get the nextjs dev server segfaulting quite a lot on node v14, but not on v16 — I'm not sure what the issue is exactly. Maybe I introduced a regression, I'm not sure.

I get segfaults all day long using node `14.19.0`. It basically segfaults on every recompile/live reload cycle.

daytonn:
> f you happen to notice a "flash of serif" on the login page, then it looks like it's down to how we're loading our webfonts, though I'm not sure how I managed to expose this as an issue.

I've noticed this as well, its not just you.

ThisIsMissEm:
> > When testing this, I did get the nextjs dev server segfaulting quite a lot on node v14, but not on v16 — I'm not sure what the issue is exactly. Maybe I introduced a regression, I'm not sure.
> 
> I get segfaults all day long using node `14.19.0`. It basically segfaults on every recompile/live reload cycle.

Good news! I've a "fix": https://twitter.com/_ijjk/status/1497390208853037057?s=21

That config change for me fixes the server segfaulting

ThisIsMissEm:
> > f you happen to notice a "flash of serif" on the login page, then it looks like it's down to how we're loading our webfonts, though I'm not sure how I managed to expose this as an issue.
> 
> I've noticed this as well, its not just you.

Oh! Okay! I moved a css component higher up in the tree because it doesn't change often & then noticed it. Perhaps an independent fix then

src/SessionReady/index.jsx
ThisIsMissEm: Ended up replacing this with just the `session` from `useSession()`

components/sessionReady/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+export default function SessionReady({ children }) {
+  const [sessionError, setSessionError] = useState();
+  const { session, sessionRequestInProgress } = useSession();
+  const router = useRouter();
+
+  const onError = useCallback(
+    (error) => {
+      setSessionError(error);
+      // FIXME: Remove once interaction_required bug in SDK is resolved:
+      session.logout();
```
ThisIsMissEm, line 39: @daytonn @VirginiaBalseiro @NSeydoux I've added a temporary `session.logout()` here whilst we wait on the fix for the interaction_required bug; Sure "logging out" of the SDK isn't great, but it's not a logout from the IDP, it just prevents the retries from heavily looping.

daytonn (approved):

VirginiaBalseiro (approved):

components/sessionReady/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+    session.on("sessionRestore", onSessionRestore);
+
+    return () => {
+      session.off("error", onError);
+      session.off("logout", onLogout);
+      session.off("sessionRestore", onSessionRestore);
+    };
+  }, [session, onError, onLogout, onSessionRestore]);
+
+  if (sessionError) {
```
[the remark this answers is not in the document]
ThisIsMissEm: @VirginiaBalseiro is there a way to trigger a toast-popup or something similar? That might be better than rendering something straight here, that way we do a toast + redirect to login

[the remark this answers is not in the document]
VirginiaBalseiro: @ThisIsMissEm Yes, we have an Alert component that you can set by using the functions in the Alert context: https://github.com/inrupt/pod-browser/blob/main/src/contexts/alertContext/index.jsx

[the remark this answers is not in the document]
ThisIsMissEm: So just `const { alertError } = useContext(AlertContext)` and then `alertError("my error message here")` right?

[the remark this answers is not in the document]
VirginiaBalseiro: Yes!

components/sessionReady/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    [router]
+  );
+
+  const onLogout = useCallback(async () => {
+    await router.push("/login");
+  }, [router]);
+
+  useEffect(() => {
+    session.on("error", onError);
```
ThisIsMissEm, line 56: Heads up folks: this is actually "onError" not "error", and that is why events should always be named with constants.

They are actually exported as constants, I just didn't know: ```import { EVENTS } from "@inrupt/solid-client-authn-core"```

ThisIsMissEm:
Okay, this is getting further along — turned out to have some bugs in the login handling, which needed a deeper fix & clean-up.

There is one known bug at the moment, which is the back behaviour in mobile safari gets stuck at the loading indicator after logging in, which could be fixed with this: https://gomakethings.com/fixing-safaris-back-button-browser-cache-issue-with-vanilla-js/

components/login/index.jsx
```diff
@@ -78,28 +104,24 @@ export default function Login() {
         alt="Inrupt PodBrowser"
         className={bem("login-form__logo")}
       />
-      <LoginButton
```
ThisIsMissEm, line 81: I've explicitly opted to not use the `LoginButton` component, as we actually have additional logic that we need to call before the login redirect happens (specifically the `persist()` call) — in the future, the SDK may have a `beforeRedirect` event, that would allow us to install this handler at a global level, rather than doing it like it's done here.

components/login/provider/__snapshots__/index.test.jsx.snap
```diff
@@ -9,38 +9,38 @@ exports[`ProviderLogin form renders a validation error if login fails 1`] = `
...
-                aria-invalid="true"
+                aria-invalid="false"
                 autocapitalize="none"
                 autocomplete="off"
                 class="PodBrowser-input PodBrowser-input PodBrowser-input PodBrowser-inputFocused PodBrowser-inputAdornedEnd PodBrowser-inputAdornedEnd"
                 id="provider-select"
+                inputmode="url"
                 spellcheck="false"
-                type="url"
+                type="text"
```
ThisIsMissEm, line 35: Whilst yes, this has changed to `"text"` note line 33, where it's now using `inputmode="url"` which is what controls the keyboard on mobile being a "url" entry keyboard; This means we don't have the browser execute validation logic on the textbox, as the value may not be a valid "url" (e.g., `inrupt.net`) but that can still be made into a valid url and validated as a IdP.

components/login/index.jsx
```diff
@@ -52,23 +55,46 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
 export const AUTH_OPTIONS = {
   clientName: CLIENT_NAME,
+  clientId: CLIENT_APP_WEBID,
 };
 
-export default function Login() {
```
ThisIsMissEm, line 57: Whilst the component is `Login` it's easily confused with `components/pages/login`, and in the place where it's used, it's called `LoginForm`, so this helps with ambiguity in the code.

components/login/provider/index.jsx
```diff
@@ -21,7 +21,7 @@
 
 /* eslint react/jsx-props-no-spreading: 0 */
 
-import React, { createRef, useEffect, useState } from "react";
```
ThisIsMissEm, line 24: I am deeply sorry for how the diff in this file ended up, it's pretty nasty, I know. Try switching to side-by-side.

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
-    loginFieldRef.current?.querySelector("input").focus();
-  }, [provider, loginFieldRef]);
+      if (reason === "blur") {
+        return;
+      }
 
-  const onProviderChange = setupOnProviderChange(setProviderIri, setLoginError);
-  const handleLogin = setupLoginHandler(login, setLoginError);
-  const onError = setupErrorHandler(setLoginError);
-  const providersWithIdp = provider ? [provider, ...providers] : providers;
```
ThisIsMissEm, line 143: I ended up removing the provider given by the `?idp=` parameter from the autocomplete options, as it has no icon and would likely look out of place in the list, and the user is probably just likely to hit "go" or enter when the page loads.

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
+                  }}
                   error={!!loginError}
                   margin="none"
                   variant="outlined"
-                  type="url"
                   aria-describedby={loginError ? "login-error-text" : null}
                   data-testid={TESTCAFE_ID_LOGIN_FIELD}
                   ref={loginFieldRef}
                   value={inputValue}
-                  onKeyUp={(e) => {
```
ThisIsMissEm, line 198: this is handled by `onChange` and `onInputChange` (specifically the "create-option" reason)

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
                   {...params}
+                  inputProps={{
+                    ...params.inputProps,
+                    autoComplete: "off",
+                    inputMode: "url",
+                  }}
                   error={!!loginError}
                   margin="none"
                   variant="outlined"
-                  type="url"
```
ThisIsMissEm, line 193: Replaced with inputMode instead, as to not have input validation from the browser here (which isn't needed as we're validating it using our own code)

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
-              id="login-error-text"
-              style={{ marginBottom: theme.spacing(1) }}
-            >
-              {getErrorMessage(loginError)}
+            <FormHelperText id="login-error-text" style={{ marginBottom: 0 }}>
+              {loginError}
             </FormHelperText>
-          ) : null}
+          ) : (
+            <p style={{ margin: 0 }}>&nbsp;</p>
```
ThisIsMissEm, line 209: Small little hack to make the page not "jump" when the single-line error message is shown.

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
+  const loginFieldRef = createRef();
 
-export function setupErrorHandler(setLoginError) {
-  return (error) => {
-    setLoginError(error);
-  };
-}
+  useEffect(() => {
+    loginFieldRef.current?.querySelector("input")?.focus();
+  }, [loginFieldRef]);
```
ThisIsMissEm, line 80: Always autofocusing seems like the right UX decision here.

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
-import getConfig from "../../../constants/config";
 
 export const TESTCAFE_ID_LOGIN_FIELD = "login-field";
 export const TESTCAFE_ID_GO_BUTTON = "go-button";
 
-const providers = getIdentityProviders();
-const hostname = getCurrentHostname();
+const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
-const CLIENT_APP_WEBID = isLocalhost(hostname)
```
ThisIsMissEm, line 59: This is handled in the parent component

components/login/provider/index.jsx
```diff
@@ -33,217 +33,203 @@ import {
...
+          Go
+        </Button>
       </div>
     </form>
   );
 }
 
 Provider.propTypes = {
-  // eslint-disable-next-line react/forbid-prop-types
-  defaultError: T.object,
```
ThisIsMissEm, line 239: This seemed to just exist for testing, as it was never passed when the component was actually used in the app — We can work on a different testing strategy for this instead.

components/pages/index/index.jsx
```diff
@@ -23,35 +23,23 @@ import { useEffect } from "react";
...
-  useRedirectIfLoggedOut();
 
   const { session } = useSession();
-  const { webId = "" } = session.info;
-  const { data: podIris = [] } = usePodIrisFromWebId(webId);
-  const [podIri] = podIris;
+  const { data: podIris = [] } = usePodIrisFromWebId(session.info.webId);
 
   useEffect(() => {
-    if (previousPage && previousPage !== "/") {
```
ThisIsMissEm, line 43: This is now handled by the login & session management

components/pages/login/index.jsx
```diff
@@ -57,22 +58,16 @@ const links = [
...
 
-  useEffect(() => {
-    if (!localStorage) return;
-    if (previousPage) {
-      localStorage.setItem("previousPage", previousPage);
-    }
-  }, [previousPage]);
+  // Redirect to root if you're already logged in:
+  if (session.info.isLoggedIn) {
+    router.replace("/");
```
ThisIsMissEm, line 69: This is only if you manually navigate to `/login` after being already logged in

components/deleteContactButton/index.test.jsx
```diff
@@ -29,7 +29,11 @@ jest.mock("@inrupt/solid-client");
 describe("Delete contact button", () => {
   test("it renders a delete contact button", () => {
     const { asFragment } = renderWithTheme(
-      <DeleteContactButton onDelete={jest.fn()} name="Test" />
+      <DeleteContactButton
+        onDelete={jest.fn()}
+        name="Test"
+        webId="https://id.inrupt.com/test"
```
ThisIsMissEm, line 35: Added `webId` here to prevent required prop warning in tests

components/login/__snapshots__/index.test.jsx.snap
```diff
@@ -16,22 +16,74 @@ exports[`Login form renders a login page with a sign in button and a 'Sign in wi
       src="/pod-spaces-logo.svg"
       width="160"
     />
+    <form>
```
ThisIsMissEm, line 19: Added the wrapping form because something weird seemed to be happening without it, it may not be needed

ThisIsMissEm:
@daytonn @VirginiaBalseiro @solid-akb this is now ready for another review, I'm just going to split that one WIP into it's three or four parts, but that shouldn't be a code change, just a commit change, so as long as you comment on the PR and not the commits it'll be fine.

It is perhaps important to note: I have not touched `/privacy` it may need an `<AuthenticatedRoute>` component in it's routes, and there is no tests at present for `<AuthenticatedRoute>` and `<AuthenticationProvider>`, nor for that the pages have a wrapping `<AuthenticatedRoute>` component (due to nextjs's lack of being able to unit test pages).

VirginiaBalseiro (approved):

daytonn (approved):
:+1:

src/authentication/AuthenticationProvider.jsx
```diff
@@ -0,0 +1,95 @@
...
+    restore();
+  }, [restore]);
+
+  const onLogout = useCallback(async () => {
+    await router.push("/login");
+  }, [router]);
+
+  useEffect(() => {
+    session.on(EVENTS.ERROR, onError);
+    // FIXME: Replace with EVENTS once solid-client-authn-js #2002 is merged
```
ThisIsMissEm, line 71: @daytonn @VirginiaBalseiro @solid-akb these can be fixed if we upgrade to the later versions of the SDK, but doing so results in a ridiculously churn to the package-lock.json: https://github.com/inrupt/pod-browser/commit/a6fdb4b722eb4bc0f549b3db2face16db1a61d0f

PR #442, opened 2022-03-01
Refactor error responder to accept unauthorized message

daytonn:
## Description
As I've been working with the address book code I noticed that there was some extra juggling necessary to respond with the correct error message using `createResponder` in `src/solidClientHelpers/utils.js`. I pulled that logic out of address book and put it in the responder code.

## Changes
- Refactored `addressBook.js` to use the refactored code
- Refactored `createResponder` to respond with an error function that can handle unauthorized requests

## Testing
- run `npm test`
- verify all tests still pass

## Commit checklist

## Interested parties
@VirginiaBalseiro @solid-akb @ajacksified 

## Notes
I'm not 100% sold on this myself. It does seem like this code does not belong in `addressBook` (with the exception of the configurable message). It does seem like we might want to handle unauthorized requests like this in the error responder because it seems silly to modify the error message provided by the thing responsible for the error message. I can see this pattern possibly getting out of hand but at that point we'd have to question the sanity of using these responders more generally if they're so diverse they defy consolidation.

I should also note that even though all the tests pass, I'm not confident that the UI remains the same. We will now be showing unauthorized errors where we were otherwise deferring to the fetch error. I _think_ this would be a good thing but I'm not confident that is the case.

### DIFF
--- a/src/addressBook/index.js
+++ b/src/addressBook/index.js
@@ -218,15 +218,10 @@ export async function saveNewAddressBook(
   { iri, owner, title = "Contacts" },
   fetch
 ) {
-  const { respond, error } = createResponder();
+  const { respond, error } = createResponder({
+    unauthorizedMessage: "You do not have permission to create an address book",
+  });
   const { response: existingAddressBook } = await getResource(iri, fetch);
-  const respondWithError = (msg) => {
-    if (isHTTPError(msg, ERROR_CODES.UNAUTHORIZED)) {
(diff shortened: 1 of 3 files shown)


solid-akb:
I like the move towards separating the errors to the utils file. 👍 Also on board with using this for now until we have more errors and then we can look at again. I think we should ask @KyraAssaad what she thinks about the new error handling.

solid-akb (approved):
Looks good

PR #446, opened 2022-03-02
Apps 1189 ensure adding contacts via web id works for all profile types

daytonn:
## Description
I began writing this PR with no small amount of confusion around how the new profiles affected PodBrowser. When I created this ticket, I didn't realize that this issue was unrelated to profiles. That being said, along the way I found some opportunities to clarify some things as I tend to make these kinds of small changes as I work through a story. This PR reflects those changes.

## Changes
- Updated the `addContact` component to take accommodate the changes
- Pulled `agentSearchForm` constants out of the component
- Changed the `value` prop on `agentSearchForm` for clarity
- Added a default empty array value for `permissions` on `agentSearchForm`
- Pulled the duplicate contact logic into a function
- Changed `filter` to `find` for checking existing contacts so `length` checking is unnecessary
- Created an `isPodOwner` utility because I saw this code duplicated in a couple places

## Testing
- Try adding a known working webID as a contact
- Verify you can add the contact

## Commit checklist

## Interested parties
@VirginiaBalseiro @solid-akb

### DIFF
--- a/components/agentSearchForm/index.jsx
+++ b/components/agentSearchForm/index.jsx
@@ -30,30 +30,46 @@ import {
   SimpleInput,
 } from "@inrupt/prism-react-components";
 import { useSession } from "@inrupt/solid-ui-react";
+import { isPodOwner } from "../../src/solidClientHelpers/utils";
 
 export const TESTCAFE_ID_ADD_AGENT_BUTTON = "add-agent-button";
 
+const AGENT_TYPE_MAP = {
+  contacts: {
+    OWN_WEBID_ERROR_MESSAGE: "You cannot add yourself as a contact.",
+  },
+  permissions: {
+    OWN_WEBID_ERROR_MESSAGE: "You cannot overwrite your own permissions.",
+  },
+};
+
+function isDuplicateContact(permissions, agentId) {
+  return permissions.filter((p) => p.webId === agentId).length;
+}
+
 export function setupSubmitHandler(
-  value,
+  agentId,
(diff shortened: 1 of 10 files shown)


VirginiaBalseiro (approved):
👍

components/agentSearchForm/index.jsx
```diff
@@ -30,30 +30,46 @@ import {
   SimpleInput,
 } from "@inrupt/prism-react-components";
 import { useSession } from "@inrupt/solid-ui-react";
+import { isPodOwner } from "../../src/solidClientHelpers/utils";
 
 export const TESTCAFE_ID_ADD_AGENT_BUTTON = "add-agent-button";
 
+const AGENT_TYPE_MAP = {
+  contacts: {
```
ajacksified, line 38: Just a style question - should `contacts` and `permissions` be capitalized (or the actual types camelcase)? AGENT_TYPE_MAP.contacts.OWN_WEBID_ERROR_MESSAGE feels a little awkward.

solid-akb (approved):

components/agentSearchForm/index.jsx
daytonn: @ajacksified I spoke too soon. It does look goofy when you write it out like that but that's not how we're using it. See https://github.com/inrupt/pod-browser/blob/f32831cfe711796b7ba2d79b0d81d39376964e68/components/agentSearchForm/index.jsx#L134

PR #447, opened 2022-03-02
fix: ignore falsy WebID values when loading available pods

ThisIsMissEm:
## Description

There was a bug where the `Home` page component would try to load a dataset located at `""` which resulted in an error being through, this happens due to `webId` having been defaulted to `""` if `session.info.webId` isn't present. In theory that shouldn't really happen, but may have still been happening due to the entire page still rendering when you're logged out — this shouldn't be an issue with #437 merged, but this helps make this code just a little more defensive before you can move across to `getPodUrlAll` from `@inrupt/solid-client` v1.20.0 (next release)

This could've been causing some weird bugs that contained complex misdirection.

Additionally doing `getSolidDataset("")` can actually result in oAuth 2.0 / OIDC URL parameters getting stashed in a dataset struct.

## Changes

- Makes `usePodIrisFromWebId` return an empty array if passed a falsy value for the `webid` argument. Arguably this could break if your WebID is "-1", "false" or similar, but given WebID's must be IRIs / URIs, I think we're fine with that.
- Adds some test coverage, even though that code will be deleted shortly.

## Commit checklist

- [x] The changelog has been updated, if applicable. (I don't think it's applicable? There does not appear to be a changelog in this repo)

## Interested parties

@inrupt/apps-developers

### DIFF
--- a/components/pages/index/__snapshots__/index.test.jsx.snap
+++ b/components/pages/index/__snapshots__/index.test.jsx.snap
@@ -1,3 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Index page Renders null if there are no pod iris 1`] = `<DocumentFragment />`;
+
+exports[`Index page Renders null if there is an empty array of pod iris 1`] = `<DocumentFragment />`;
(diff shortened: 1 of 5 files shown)


solid-akb (approved):
Awesome updates - great catch figuring out the root cause of the bug and the *other* implications of passing in an empty string.

PR #451, opened 2022-03-14
Small sharing cleanup

solid-akb:
## Description
I broke up the PR I am currently working on around re-renders in the Sharing Panel into refactor code and items that will cause change to data flow. This PR is around refactoring so there isn't just one big one when I get the re-render figured out. 
## Changes
- if there is something that is an array, if it's not set or defined yet, instead of defaulting to null we default to an empty array. This is nice because we don't change the type at any point and also so that things are less prone to break. If the array is empty we just cycle through an empty array instead of the code breaking on null. A big gotcha (that I covered for the permissions but we'll need to look at elsewhere as we move to this appraoch) is that [] is truthy so we will have to protect against an empty array by calling `if (!someValue.length)` instead of `if (!someValue)` . This goes the same for strings starting as `""` instead of `null` but this will break less because "" is falsey. 
- fixed bug where we thought we were getting an array of profiles from SDK (even if there was just one) 
- took out some filtering and sorting functions from the component and moved them to a util file. 
- commented out code that is not currently being used that was bringing test coverage down
- random cleanup of extra whitespace
- move style related variables to the top of the component
## Testing
Tests have been updated. This PR has no change to the end user. 
## Commit checklist

- [NA ] All acceptance criteria are met.
- [ x] Includes tests to ensure functionality and prevent regressions.

## Interested parties
@VirginiaBalseiro  @daytonn @ajacksified


components/resourceDetails/utils.js
```diff
@@ -0,0 +1,58 @@
...
+import {
+  PUBLIC_AGENT_PREDICATE,
+  PUBLIC_AGENT_TYPE,
+} from "../../src/models/contact/public";
+
+export function filterPermissionsByType(permissions, type) {
+  return permissions?.filter((permission) => permission.type === type) || [];
+}
+
+export function isPublicAgentorAuthenticatedAgentType(type) {
```
daytonn, line 35: We should capitalize the "or": `isPublicAgentOrAuthenticatedAgentType`

daytonn, line 35:
```suggestion
export function isPublicAgentOrAuthenticatedAgentType(type) {
```

daytonn (approved):

components/resourceDetails/resourceSharing/agentAccessList/__snapshots__/index.test.jsx.snap
```diff
@@ -35,24 +35,8 @@ exports[`AgentAccessList renders an empty list of permissions 1`] = `
...
-      <circle
-        class="PodBrowser-circle PodBrowser-circleIndeterminate"
-        cx="44"
-        cy="44"
-        fill="none"
-        r="20.2"
-        stroke-width="3.6"
-      />
-    </svg>
+  <div>
```
VirginiaBalseiro, line 38: This snapshot should be rendering a loading spinner. The spinner does appear in the vercel deployment so it would appear it's something to do with the test itself.

src/contexts/accessControlContext/index.jsx
```diff
@@ -24,7 +24,7 @@ import T from "prop-types";
 
 const defaultValues = {
   accessControl: null,
-  accessControlType: null,
+  accessControlType: "",
```
VirginiaBalseiro, line 27: this one should be null, since the value is set from the return value of the `useAccessControlType` hook, and we check whether that value is falsy in the `useAccessControl` hook.

VirginiaBalseiro (changes requested):
Overall looks good! Left two comments. In addition, when testing, I get an error when trying to update permissions, and the panels are not displayed (endless spinner): 

image

src/contexts/accessControlContext/index.jsx
VirginiaBalseiro: It doesn't necessarily mean that there is not an access control type, it just means that for whatever reason we couldn't determine it. The SDK methods used return booleans, and we check for an ACR or an ACL.

VirginiaBalseiro: In addition, f we're updating the default values here we will need to update it in the corresponding hook and related code and docs as well.

components/resourceDetails/utils.js
solid-akb: Didn't know the Suggested Change feature existed in GH. Super helpful. Just accepted and committed. 👍

src/contexts/accessControlContext/index.jsx
VirginiaBalseiro: > was successful but the answer is that the user doesn't have access control instead of a null kind of leaves me guessing, was the call ever made?

We do check for the existence of either an error or a response to see if the hook is still validating so that could potentially answer that question.

VirginiaBalseiro:
When running locally and trying to open the side panel: 

image

VirginiaBalseiro:
When running locally and trying to open the side panel:

image

VirginiaBalseiro:
> Overall looks good! Left two comments. In addition, when testing, I get an error when trying to update permissions, and the panels are not displayed (endless spinner):
> 
> image

I don't see this behavior anymore when running locally so might have been something with the Vercel deployment?

components/resourceDetails/utils.js
```diff
@@ -0,0 +1,58 @@
...
+
+export function sortByAgentName(permissions) {
+  return permissions.sort((a, b) => {
+    return a.profile?.name?.localeCompare(b.profile?.name);
+  });
+}
+
+export function preparePermissionsDataForTable(data) {
+  const publicAndAuthPermissions = data?.filter((p) =>
+    isPublicAgentorAuthenticatedAgentType(p.type)
```
VirginiaBalseiro, line 53: this needs updating too to capitalize the "or". Everything else looks good!

solid-akb: I updated this in github so it should be fixed if you `git pull`

solid-akb: 👍

components/resourceDetails/resourceSharing/agentAccessList/index.test.jsx
```diff
@@ -82,7 +82,7 @@ describe("AgentAccessList", () => {
   useFetchProfile.mockReturnValue({ webId });
 
   it("renders loading while loading permissions for access control", () => {
-    usePermissions.mockReturnValue({ permissions: null });
+    usePermissions.mockReturnValue({ permissions: [] });
```
VirginiaBalseiro, line 85: AgentAccessList (the component) has this line: 

L44:  const loading = !permissions;

which checks is permissions is null, which is causing this snapshot to fail. We need to update that to check for length.

PR #454, opened 2022-03-19
Removed forbid prop types

solid-akb:
## Description
We used `eslint-disable-next-line react/forbid-prop-types` pretty gratuitously throughout the code. Removed the instance when it wasn't protecting anything and when we used T.Object. From now on we should use `T.oneOfType([Object])`
## Changes
Deleted extra `eslint-disable-next-line react/forbid-prop-types` and replaced T.Object with `T.oneOfType([Object])` which does not throw a warning.
## Testing
NA
## Commit checklist

- [ NA] All acceptance criteria are met.
- [ NA] Includes tests to ensure functionality and prevent regressions.
- [ NA] The changelog has been updated, if applicable.

## Notes
Use `T.oneOfType([Object])` when you want to pass in an Object as a prop.

### DIFF
--- a/__testUtils/testApp.jsx
+++ b/__testUtils/testApp.jsx
@@ -49,6 +49,5 @@ TestApp.defaultProps = {
 
 TestApp.propTypes = {
   children: T.node.isRequired,
-  // eslint-disable-next-line react/forbid-prop-types
-  session: T.object,
+  session: T.oneOfType([Object]),
 };
--- a/__testUtils/withTheme.jsx
+++ b/__testUtils/withTheme.jsx
@@ -36,8 +36,7 @@ export default function WithTheme(props) {
 }
 
 WithTheme.propTypes = {
-  // eslint-disable-next-line react/forbid-prop-types
-  theme: T.object,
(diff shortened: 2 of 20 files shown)


daytonn (approved):
I love it. However, do we want to just disable this rule? It seems like the main value here is to provide specific type information for non-scalar values. In this case, we're essentially saying "no thanks" but still bending this rule to get away with essentially ignoring this rule.

We could just add `"react/forbid-prop-types": 0` to our eslint config and just get rid of the comments.

Either way I'm in favor of this change.

solid-akb:
> I love it. However, do we want to just disable this rule? It seems like the main value here is to provide specific type information for non-scalar values. In this case, we're essentially saying "no thanks" but still bending this rule to get away with essentially ignoring this rule.
> 
> We could just add `"react/forbid-prop-types": 0` to our eslint config and just get rid of the comments.
> 
> Either way I'm in favor of this change.

There were a few places where it felt like it could be helpful and/or I couldn't find a quick fix so I didn't nuke it entirely. Let's have the convo about how to handle those.

PR #456, opened 2022-03-22
Feat: add research survey widget

VirginiaBalseiro:
## Description

This PR adds a button to the login page which opens a widget containing a link to the PodBrowser research survey.

## Changes

New feature added to the login page as describes above

## Testing

- Navigate to login page and find the button at the bottom of the page (test mobile and desktop according to acceptance criteria)
- Test widget displays properly, link works as intended, and close button/outside click closes the widget
- Test it is keyboard accessible

## Commit checklist

### DIFF
--- a/components/surveyWidget/widgetBox/index.jsx
+++ b/components/surveyWidget/widgetBox/index.jsx
@@ -41,7 +41,7 @@ export default function WidgetBox({ id, open, anchorEl, onClose }) {
   return (
     <Popover
       data-testid={TESTCAFE_ID_WIDGET_BOX}
-      classes={{ paper: classes.widgetBox }}
+      classes={{ paper: classes["survey-widget-box"] }}
       id={id}
       open={open}
       anchorEl={anchorEl}
--- a/components/surveyWidget/widgetBox/index.jsx
+++ b/components/surveyWidget/widgetBox/index.jsx
@@ -1,0 +1,93 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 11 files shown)


solid-akb (commented):
Do we want the changes in the package-lock file? These don't feel necessary or we should all have the same prettier config if we don't already.

VirginiaBalseiro:
> Do we want the changes in the package-lock file? These don't feel necessary or we should all have the same prettier config if we don't already.

Oh I missed this! I changed my config for another project and forgot to switch back. Will revert and push in a min.

components/surveyWidget/widgetBox/index.jsx
```diff
@@ -0,0 +1,93 @@
...
+      <span className={bem("survey-widget-close-button-wrapper")}>
+        <button
+          type="button"
+          className={bem("survey-widget-close-button")}
+          onClick={onClose}
+          data-testid={TESTCAFE_ID_CLOSE_WIDGET_BUTTON}
+        >
+          <Icons name="cancel" />
+        </button>
+      </span>
```
daytonn, line 66: Spans are inline elements and should only be used to wrap text. Divs are basically spans meant for stuff other than inline-text.

```suggestion
      <div className={bem("survey-widget-close-button-wrapper")}>
        <button
          type="button"
          className={bem("survey-widget-close-button")}
          onClick={onClose}
          data-testid={TESTCAFE_ID_CLOSE_WIDGET_BUTTON}
        >
          <Icons name="cancel" />
        </button>
      </div>
```

components/surveyWidget/widgetBox/styles.js
```diff
@@ -0,0 +1,79 @@
...
+import { createStyles } from "@solid/lit-prism-patterns";
+
+const BOTTOM_POSITION_CLOSE_BUTTON = "88%";
+const LEFT_POSITION_CLOSE_BUTTON = "81%";
+const SHADOW = "0 6px 23px 0 rgba(149, 149, 149, 0.2)";
+const DARK_TEAL_TEXT_COLOR = "073D6C";
+
+export default function styles(theme) {
+  return createStyles(theme, ["button"], {
+    widgetBox: {
```
daytonn, line 30: nitpick: I notice all the other keys are hyphenated. We should keep these consistent. I would personally prefer using camel case to avoid all the quoting but it doesn't really matter either way.

daytonn (approved):
Small nitpicks but otherwise :+1:

components/surveyWidget/widgetBox/styles.js
VirginiaBalseiro: The reason for this is how we use the classes to override MUI vs bem, but I switched it to hyphenated for consistency.

PR #457, opened 2022-03-23
fix minimist vulnerability via npm audit fix

VirginiaBalseiro:
Fix minimist vulnerability via npm audit fix.

There are also unrelated linting changed in the lock file which happen on the latest npm when running npm install.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2313,9 +2313,13 @@
       "version": "12.1.0",
       "resolved": "https://registry.npmjs.org/@next/swc-android-arm64/-/swc-android-arm64-12.1.0.tgz",
       "integrity": "sha512-/280MLdZe0W03stA69iL+v6I+J1ascrQ6FrXBlXGCsGzrfMaGr7fskMa0T5AhQIVQD4nA/46QQWxG//DYuFBcA==",
-      "cpu": ["arm64"],
+      "cpu": [
+        "arm64"
+      ],
       "optional": true,
-      "os": ["android"],
+      "os": [
+        "android"
+      ],
       "engines": {
         "node": ">= 10"
       }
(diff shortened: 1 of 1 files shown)


solid-akb:
I think the bigger question on this PR is do we all have the same config so that our package-lock files are consistent. I am still running into issues and I think @ThisIsMissEm also had a 30k+ diff last time she tried to PR to PB. @daytonn have you been able to replicate a package-lock file that matches main?

ThisIsMissEm:
I was getting a diff on main just running `npm install`, looks like maybe the package-lock.json was accidentally run through prettier maybe.

As for the huge diff, that's why I'd suggested I look into it, as I'm not sure why all the build toolchain is being updated here, it's quite strange.

I suspect it may be because of a bad dep in the SDK that we fixed, but that shouldn't have been keeping you stuck at an older version of babel and such, I don't think.

It may be an idea to merge the switch to next & then do the upgrade, as that removes more of your build toolchain.

Emelia

> On 23. Mar 2022, at 17:25, Anna Kallenborn-Bolden ***@***.***> wrote:
> 
> ﻿
> I think the bigger question on this PR is do we all have the same config so that our package-lock files are consistent. I am still running into issues and I think @ThisIsMissEm also had a 30k+ diff last time she tried to PR to PB. @daytonn have you been able to replicate a package-lock file that matches main?
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.

-- 
This e-mail, and any attachments thereto, is intended only for use by the 
addressee(s) named herein and may contain legally privileged, confidential 
and/or proprietary information. If you are not the intended recipient of 
this e-mail (or the person responsible for delivering this document to the 
intended recipient), please do not disseminate, distribute, print or copy 
this e-mail, or any attachment thereto. If you have received this e-mail in 
error, please respond to the individual sending the message, and 
permanently delete the email.

solid-akb (approved):

PR #458, opened 2022-03-23
remove unnecessary await before act

VirginiaBalseiro:
This PR removes `await` when using act in tests in the `useReturnUrl` hook because it was causing a console warning. 

`act` cannot be awaited unless used in its sync version: https://reactjs.org/docs/testing-recipes.html#act


solid-akb (approved):
Awesome thank you!

PR #460, opened 2022-03-23
Add filter to redact certain Sentry logs

daytonn:
## Description
This PR adds functionality to redact certain breadcrumbs that include sensitive information.

## Changes
- Added a `sentryFilter` module to contain Sentry log filters
- Added a `listHelpers` module and added a partition function

## Testing
- It's complicated

## Commit checklist

## Interested parties
@solid-akb @VirginiaBalseiro 

## Notes
This one is a bit harder to test than normal since we'll need to trigger the offending errors and then check Sentry to ensure that they've been redacted.

### DIFF
--- a/src/sentryFilter/index.js
+++ b/src/sentryFilter/index.js
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+function hasBreadcrumbs(event) {
+  return !!(event && event?.breadcrumbs && event?.breadcrumbs?.length);
+}
+
+export function redactCodeParameter(string) {
+  return string.replace(/([?&])(code=[^&]+)/, "$1code=[REDACTED]");
+}
+
+function redactBreadcrumbs(crumbs) {
+  return crumbs.map((crumb) => {
+    if (crumb.category !== "navigation") {
+      return crumb;
+    }
+
+    if (crumb.from) {
+      Object.assign(crumb, { from: redactCodeParameter(crumb.from) });
+    }
+
+    if (crumb.to) {
+      Object.assign(crumb, { to: redactCodeParameter(crumb.to) });
+    }
+    return crumb;
+  }, []);
+}
+
+export function filterSentryEvent(event) {
+  if (!hasBreadcrumbs(event)) {
+    return event;
+  }
+
+  const breadcrumbs = redactBreadcrumbs(event.breadcrumbs);
+
+  return { ...event, breadcrumbs };
+}
+
+export default { filterSentryEvent };
--- a/src/sentryFilter/index.test.js
+++ b/src/sentryFilter/index.test.js
@@ -1,0 +1,125 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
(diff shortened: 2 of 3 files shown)


solid-akb (commented):
Do you have a screenshot where you've added sensitive data and then it's redacted in sentry? I totally see what you're saying how it's hard to test, so happy to just look at a screenshot to see the difference. I can't totally tell what sensitive data is being removed from the logs. What does this filter? Also, I wonder if there is a simpler way to do this?

daytonn:
@solid-akb This specifically filters out `code=xxx` parameters logged in `breadcrumbs.navigation` objects sent to Sentry. We need to trigger an error condition that has a code redirect path.

@ThisIsMissEm also pointed out that there is a ticket on the Sentry side for filtering sensitive data.

> There’s also an open support ticket with Sentry to ensure from their side that they scrub this data (they should’ve already been), it’s #61981

The following is the error that triggered this ticket:
https://sentry.io/organizations/inrupt/issues/3081427818/?frontend_events=%7B%22event_name%22%3A%22Sign+Up%22%2C%22event_label%22%3A%22google%22%7D&project=5383681&query=is%3Aunresolved&sort=date&statsPeriod=14d#tags

src/sentryFilter/index.js
```diff
@@ -0,0 +1,58 @@
...
+export function containsSensitiveInformation(string) {
+  return !!string.match(/code=.+\b/);
+}
+
+export function redactCodeParameter(string) {
+  return string.replace(/(.+)?(code=[A-z,0-9]+)(\b)?/, "$1code=[REDACTED]$3");
+}
+
+export function redactNavigationCrumbs(crumbs) {
+  return crumbs.reduce((acc, crumb) => {
```
ThisIsMissEm, line 37: this would be better written using `map` as that's what you're doing.

src/sentryFilter/index.js
```diff
@@ -0,0 +1,58 @@
...
+    if (crumb.to) {
+      Object.assign(crumb, { to: redactCodeParameter(crumb.to) });
+    }
+    return [crumb, ...acc];
+  }, []);
+}
+
+export function filterCodeRedirects(event) {
+  if (!hasBreadcrumbs(event)) return event;
+  const [navCrumbs, otherCrumbs] = partition(event?.breadcrumbs);
```
ThisIsMissEm, line 51: I'm not sure why you need `partition` here when a simple `map` would work fine?

```
event.breadcrumbs = event.breadcrumbs.map((crumb) {
  if (crumb.category !== 'navigation') {
    return crumb;
  }

  if (crumb.to) {
    crumb.to = redactCodeParameter(crumb.to);
  }
  
  if (crumb.from) {
    crumb.from = redactCodeParameter(crumb.from);
  }

  return crumb;
});

return event;
```

Seems like it'd be a simpler implementation, but maybe I'm missing something? There's no need to be non-mutative here.

src/sentryFilter/index.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import { partition } from "../enumerable";
+
+function hasBreadcrumbs(event) {
+  return !!(event && event?.breadcrumbs && event?.breadcrumbs?.length);
+}
+
+export function containsSensitiveInformation(string) {
+  return !!string.match(/code=.+\b/);
```
ThisIsMissEm, line 29: You can do away with this regex here
```suggestion
  return string.contains("code=");
```

The regex may also potentially have unbounded complexity:
Screenshot 2022-03-25 at 14 50 03

ThisIsMissEm: (looked at that here: http://xenon.stanford.edu/~xusch/regexp/ )

src/sentryFilter/index.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export function redactCodeParameter(string) {
+  return string.replace(/(.+)?(code=[A-z,0-9]+)(\b)?/, "$1code=[REDACTED]$3");
```
ThisIsMissEm, line 33: Would it perhaps be better to parse using `URLSearchParams` rather than regexes?

src/sentryFilter/index.js
ThisIsMissEm: Also, your `partition` function takes two parameters, yet you only supply one here, so I don't think it's really partitioning between navCrumbs and non-navCrumbs — but as I mention above, I'd argue you don't actually need the `partition` in the first place.

src/sentryFilter/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+describe("filterCodeRedirects", () => {
+  it("does nothing when there is no data to filter", () => {
+    const event = {
+      breadcrumbs: [{ category: "test" }],
+    };
+
+    expect(filterCodeRedirects(event)).toMatchObject(event);
+  });
+
+  it("redacts codes sent in navigation breadcrumbs", () => {
```
ThisIsMissEm, line 71: It'd be a good idea to have a test case that checks multiple breadcrumbs, e.g., navigation, event, etc to ensure non-navigation crumbs aren't redacted/filtered

ThisIsMissEm (commented):
I would highly suggest changing this to reduce complexity, as it can definitely be done simpler (it's okay to mutate data here)

src/sentryFilter/index.js
daytonn: Whoops, I started this with the intent of filtering out navigation crumbs and ended up filtering them upstream. Good catch.

src/sentryFilter/index.js
daytonn: I _think_ you mean `includes` here but you're right. I forgot about `String.includes`.

src/sentryFilter/index.js
daytonn: I thought about going that route. I'm not opposed to it. I'm not sure how much reuse this function will get but that was the idea.

src/sentryFilter/index.test.js
daytonn: :+1:

solid-akb (approved):
looks great!!

PR #463, opened 2022-03-30
use expiration date from access request

VirginiaBalseiro:
## Description

In this PR, we use the expiration date from the access request instead of defaulting to Forever without checking if an expiration date is available.

## Testing

- Go to [test-app-access-grants-inrupt.vercel.app](https://test-app-access-grants-inrupt.vercel.app/) and generate a request (it will have an expiration date)
- Grab the URL for the request and append to the path for the URL of the Vercel deployment for this PR
- Verify a date is displayed (today's - the access request is for a few minutes)

## Commit checklist


components/consentRequestForm/__snapshots__/index.test.jsx.snap
```diff
@@ -140,7 +140,7 @@ exports[`Consent Request Form Renders a consent request form with multiple purpo
           placeholder="Consent expiry date"
           readonly=""
           type="text"
-          value="Forever"
+          value="June 23, 2022"
```
ThisIsMissEm, line 143: Careful here, unless you're [mocking time](https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers). (this is now default in [Jest 27](https://jestjs.io/blog/2021/05/25/jest-27#:~:text=Another%20default%20that%20we%20are%20changing%20affects%20Fake%20Timers%20aka%20Timer%20Mocks))

components/consentRequestForm/index.jsx
```diff
@@ -207,10 +209,11 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
...
     setDatepickerOpen(false);
   };
-
   useEffect(() => {
     if (expirationDate) {
       setSelectedDate(new Date(expirationDate));
+    } else {
+      setDateForever();
     }
   }, [expirationDate]);
```
ThisIsMissEm, line 218: I don't think you actually need this, as `expirationDate` is set when `selectedDate` is Initialised (L102-104), and as `consentRequest` comes from context, if the consent request changes, this whole component will re-render, so `expirationDate` will never change.

You could maybe shift the date parsing logic down to the DateInput and have it just return a string for the date, either "Forever" or a [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formatted date string (I assume that's what the VC spec would call for).

That way you don't need to marshal the date value here and you don't have multiple setState's for the expiration.

Also, `selectedDate` isn't actually read anywhere in the `handleSubmit` function, I assume it's meant to be.

components/consentRequestForm/index.jsx
```diff
@@ -286,10 +289,7 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
           {`${agentName || agentWebId} will have access until`}
         </span>
         <DateInput
-          // selectedDate={selectedDate}
-          // setting date to forever for now until we add functionality to change this
-          selectedDate={null}
-          setSelectedDate={setSelectedDate}
+          selectedDate={selectedDate}
           datepickerOpen={datepickerOpen}
           setDatepickerOpen={setDatepickerOpen}
```
ThisIsMissEm, line 294: this is really internal state being leaked to the parent component, I think, as the date picker automatically closes when you click outside of it; you could also have the `handleDateChange` prop actually be executed in the DateInput like:

```
const onDateChange = useCallback((newDate) => {
  setDatepickerOpen(false)
  handleDateChange(/* format(newDate as iso8601) */);
}, [setDatepickerOpen, handleDateChange]);

const onSelectForever = useCallback(() => {
    setDatepickerOpen(false)
    handleDateChange('Forever');
}, [setDatepickerOpen, handleDateChange]);
```

That way here all you have is the `handleDateChange` method which receives either `'Forever'` or an ISO 8601 date string.

ThisIsMissEm (commented):
A few comments that could improve the clarity of this code, it's mostly about keeping state local and making sure side-effects are meaningful. You may find the links I recently shared in #interest-frontend on slack interesting.

components/consentRequestForm/__snapshots__/index.test.jsx.snap
VirginiaBalseiro: What do you mean? This value is set from the expiration date from the mock consent request.

ThisIsMissEm: Ah! So it is! I looked at the mockSignedVc request accidentally last night 🤦🏻‍♀️

So if I understand the diff correctly, the date wasn't previously being used so defaulted to Forever, now it is, so it's changed the snapshot to reflect?

ThisIsMissEm: Aha, okay! Looks good then. My other comments are just ways you could improve the code

ThisIsMissEm (approved):

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -41,38 +40,29 @@ export const TESTCAFE_ID_DATE_PICKER_CALENDAR_BUTTON =
...
 export default function DateInput(props) {
-  const {
-    selectedDate,
-    setSelectedDate,
-    datepickerOpen,
-    setDatepickerOpen,
-    handleDateChange,
-    setDateForever,
-  } = props;
+  const { selectedDate, handleDateChange, setDateForever } = props;
```
ThisIsMissEm, line 43: doesn't the linter tell you to do this in the function declaration?

```
export default function DateInput({ selectedDate, handleDateChange }) {
```

ThisIsMissEm: Note, I removed `setDateForever` as `handleDateChange` should just return a string

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -41,38 +40,29 @@ export const TESTCAFE_ID_DATE_PICKER_CALENDAR_BUTTON =
...
 
-  useEffect(() => {
-    if (expirationDate) {
-      setSelectedDate(expirationDate);
-    }
-  }, [expirationDate, setSelectedDate]);
+  useOutsideClick(clickRef, () => setDatepickerOpen(false));
 
   return (
     <div ref={clickRef} className={bem("date-container")}>
```
ThisIsMissEm, line 54: I think you want this `clickRef` on `<DatePicker` not here.

components/consentRequestForm/dateInput/index.jsx
VirginiaBalseiro: Huh I thought I changed this... I will change now.

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -81,7 +71,7 @@ export default function DateInput(props) {
         classes={{ root: bem("date-button") }}
         type="button"
         // temporarily disabling until functionality to change date is available
-        disabled
+        // disabled
         aria-label="Set expiry date"
         onClick={() => setDatepickerOpen(!datepickerOpen)}
```
ThisIsMissEm, line 76: I'd recommend lifting this up to `InputBase` via `endAdornment` so it's just an Icon, not a button — https://v4.mui.com/api/input-base/#inputbase-api

That way clicking on the date will open the calendar too.

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -81,7 +70,7 @@ export default function DateInput(props) {
         classes={{ root: bem("date-button") }}
         type="button"
         // temporarily disabling until functionality to change date is available
-        disabled
+        // disabled
```
ThisIsMissEm, line 73:
```suggestion
```

components/consentRequestForm/dateInput/index.jsx
```diff
@@ -98,8 +87,6 @@ export default function DateInput(props) {
             orientation="portrait"
             variant="static"
             disablePast
-            // temporarily disabling until functionality to change date is available
-            disabled
             format="MM/dd/yyyy"
             margin="normal"
             value={selectedDate}
```
ThisIsMissEm, line 92: Not sure you wanna pass `selectedDate` wholesale here, as it may be `null`

components/consentRequestForm/index.jsx
```diff
@@ -118,6 +116,8 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
           requestor,
           access: selectedAccess.accessModes,
           resources: selectedResources,
+          expirationDate: new Date(selectedDate),
```
ThisIsMissEm, line 119: You'll want to conditionally set this, I think, so perhaps move this object up into a `const approvalOverrides` object, and then `if (selectedDate !== null) { approvalOverrides.expirationDate = new Date(selectedDate); }`

ThisIsMissEm: I guess it depends on what "Forever" means, does it mean Infinity, or does it mean "whatever the requestor wanted originally"

components/consentRequestForm/dateInput/index.jsx
ThisIsMissEm: @VirginiaBalseiro maybe do it on the parent div then?

PR #464, opened 2022-04-01
Remove conf dialog context

solid-akb:
## Description
Moving the Confirmation Dialog logic from Global Context to being managed at the Parent component level. This means that instead of changing state on ConfirmationDialogContext we are handling open/close of the Dialog in the Parent's state and the rest of the details (Title, content, onConfirm, onCancel, etc) are being passed in as props. 

It's worth noting that I only changed the Confirmation Dialogs that were straight forward. I will slowly get to the others, but in the case of Context, any time we take it out we are potentially saving re-renders so I am hoping this helps. This also serves as an example for how we can use the component moving forward. 

Because the old Confirmation Dialog that uses Context still exists, we now have two ConfirmationDialog Components. ConfirmationDialog is the Context based one that will be deprecated (along with ConfirmationDialogContext) and ConfirmationDialogProps is the new one (it will be renamed to ConfirmationDialog once the old one is deprecated, but this is the name for now to indicate that this one takes props). 
## Changes
See above
## Testing
All impacted Components have been tested.
## Commit checklist

## Notes
Why this PR and why now? This article was the inspiration https://www.zhenghao.io/posts/react-rerender


VirginiaBalseiro:
Tested delete button and works great. Unfortunately when trying to test the consent request form I get:

image


Also, minor, but maybe renaming to `ConfirmationDialogWithProps` might be a bit clearer?

solid-akb:
> Tested delete button and works great. Unfortunately when trying to test the consent request form I get:
> 
> image
> 
> The console doesn't show anything. Were you able to test this without issues?
> 
> Also, minor, but maybe renaming to `ConfirmationDialogWithProps` might be a bit clearer?

Might be an issue with the previous PR around consent - will look at it now. 

Will change to ConfirmationDialogWithProps. The idea is that it will replace ConfirmationDialog and the "WithProps" will be dropped soon, but in the meantime will update.

solid-akb:
Did you use the new Consent flow or the existing consent flow when you ran into the error? @VirginiaBalseiro

VirginiaBalseiro (changes requested):
Managed to test with a fresh consent request. There's some funky behavior: 
- When accepting the request without selecting any purposes, it sometimes displays the dialog and sometimes doesn't. 
- After denying a request, if I go back to the request and try to accept it, it displays the "deny" confirmation dialog but accepts the request nevertheless.

VirginiaBalseiro (approved):
Ignore my previous comment: it looks like at least one these behaviors is present in prod so the others might be related. Fixing in my current ticket.

PR #466, opened 2022-04-04
fix: ignore errors thrown from localstorage when persisting the returnTo URL

ThisIsMissEm:
## Description

I didn't realise `localStorage.setItem` could throw, but apparently it can, so this makes that error silent. This fixes POD-BROWSER-2D9 on Sentry.

## Changes

Just adds a try/catch

## Testing

Additional test case has been added.

## Commit checklist

- [ ] The changelog has been updated, if applicable. — Is this really changelog worthy?

### DIFF
--- a/src/authentication/useReturnUrl.js
+++ b/src/authentication/useReturnUrl.js
@@ -29,7 +29,12 @@ const useReturnUrl = () => {
 
   const persist = useCallback(() => {
     if (router.query.returnTo && router.query.returnTo.startsWith("/")) {
-      localStorage.setItem(RETURN_TO_PAGE_KEY, router.query.returnTo);
+      try {
+        localStorage.setItem(RETURN_TO_PAGE_KEY, router.query.returnTo);
+      } catch {
+        // silently fail if persisting the return to URL fails, it sucks but
+        // there isn't really a "good" user experience to have here.
+      }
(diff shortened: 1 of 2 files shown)


solid-akb (commented):
Do we want to add any error handling for this for 2.0 release? @KyraAssaad

ThisIsMissEm:
@KyraAssaad this happens in different cases, but is likely to be pretty rare, e.g., someone has changed security settings to block all localstorage or localstorage is full. 

In either case, nothing is going to work correctly as authentication via solid-client-authn-browser requires localstorage. The error handler from login should catch this as login break entirely in that case. Maybe there we could add better error handling, but it's more something we could handle in a support article.

In this specific case, it's just a small safety check that we can add that just downgrades them from automatically being redirected back to the page they requested, to redirecting them back to the root of their pod, if for some reason, authentication succeeded.

ThisIsMissEm:
@KyraAssaad specifically, this change doesn't concern itself with handling errors from all localstorage, just with the `returnTo` handling, I've separately mentioned that this could break within the SDK team, we'd likely short-circuit this error scenario in the SDK and return back an error event with a message indicating we couldn't interact with localstorage.

solid-akb (approved):

PR #467, opened 2022-04-04
changed PT to object and added eslint disable

solid-akb:
## Description
Prop types that were Object changed back to oneOfType([Object]) because oneOfType([Object]) was breaking unit tests (and is incorrect). Revisit react/forbid-prop-types comments at a later time. 
## Commit checklist

### DIFF
--- a/src/contexts/consentRequestContext/index.jsx
+++ b/src/contexts/consentRequestContext/index.jsx
@@ -44,7 +44,8 @@ function ConsentRequestProvider({
 
 ConsentRequestProvider.propTypes = {
   children: T.node.isRequired,
-  consentRequest: T.oneOfType([Object]),
+  // eslint-disable-next-line react/forbid-prop-types
+  consentRequest: T.object,
   setConsentRequest: T.func,
 };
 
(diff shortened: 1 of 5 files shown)


src/contexts/consentRequestContext/index.jsx
```diff
@@ -44,7 +44,8 @@ function ConsentRequestProvider({
 
 ConsentRequestProvider.propTypes = {
   children: T.node.isRequired,
-  consentRequest: T.oneOfType([Object]),
+  // eslint-disable-next-line react/forbid-prop-types
+  consentRequest: T.object,
```
ThisIsMissEm, line 48: Ideally this would be a `T.shape` with the properties required tested, though, tbh, this is a lot nicer handled in typescript, where you don't need runtime propTypes, because it happens via the language

solid-akb: Yes - totally agree - would love to get to T.shape but maybe something for the future?

ThisIsMissEm (approved):
This works for me and fixes the errors on `main`

src/contexts/consentRequestContext/index.jsx
ThisIsMissEm: @solid-akb oh, absolutely!

PR #468, opened 2022-04-04
Simplify CI configuration

ThisIsMissEm:
## Description

Node.js v12 is now end of life, and we only deploy to node v14; Also uses explicit steps for test and lint, rather than combining them

## Changes

I've documented anything specific in the "files" tab via comments.

## Testing

All tests ran fine

## Commit checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,24 +6,13 @@ env:
   CI: true
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
-        os: [ubuntu-18.04]
-        node-version: [14.x, 12.x]
+    runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v2-beta
-      with:
-        node-version: ${{ matrix.node-version }}
-    - name: Cache node modules
-      uses: actions/cache@v1.1.2
-      env:
-        cache-name: cache-node-modules
-      with:
-        path: node_modules
-        key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-    - run: npm ci
-    - run: npm run ci
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: npm
+      - run: npm ci
+      - run: npm run lint
+      - run: npm run test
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -6,24 +6,13 @@ env:
...
-      with:
-        path: node_modules
-        key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-    - run: npm ci
-    - run: npm run ci
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: npm
```
ThisIsMissEm, line 15: doing `cache: npm` will cache the npm cache, not `node_modules` but the `npm ci` will already have all the files and the install will be quick.

.github/workflows/ci.yml
```diff
@@ -6,24 +6,13 @@ env:
...
-        node-version: [14.x, 12.x]
+    runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v2-beta
-      with:
-        node-version: ${{ matrix.node-version }}
-    - name: Cache node modules
-      uses: actions/cache@v1.1.2
```
ThisIsMissEm, line 22: I could've upgraded to `actions/cache@v2` but `actions/cache@v3` requires node 16, so we couldn't use that. Using `cache: npm` should be good enough though for now.

package.json
```diff
@@ -7,7 +7,6 @@
     "audit": "npm audit --audit-level=moderate",
     "audit-licenses": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
     "build": "next build && next export",
-    "ci": "npm run lint && npm run test",
```
ThisIsMissEm, line 10: I'd never recommend naming a command in a way that conflicts with a built-in npm command — in this case `npm ci` exists, so I wouldn't suggest having a `npm run ci` command; By splitting these out, we also have clearer test output.

ajacksified: If nobody's using this, it's fine (`npm run ci` was a habit I'd gotten into before `npm ci` existed.) We should have a discussion about having a simple command to run audit/lint/test in localhost, but as husky runs lint and test it's not a problem for now. We've discussed removing husky elsewhere so that probably isn't the right answer long term.

solid-akb (commented):
Looks good! Will let Jack and Virginia take a look before approving.

solid-akb (approved):
👍

PR #469, opened 2022-04-04
Fix: upgrade sentry integration

ThisIsMissEm:
## Description

The modules for the sentry integration were outdated, and it turns out there's now a specific nextjs configuration that exists, so we were able to move to that relatively easily.

## Testing

Have run unit tests, and will also test that the app functions as expected on vercel

## Commit checklist


next.config.js
```diff
@@ -19,50 +19,28 @@
...
+  // Additional config options for the Sentry Webpack plugin. Keep in mind that
+  // the following options are set automatically, and overriding them is not
+  // recommended:
+  //   release, url, org, project, authToken, configFile, stripPrefix,
+  //   urlPrefix, include, ignore
 
-    return config;
-  },
+  // Suppresses all logs
+  silent: true,
```
solid-akb, line 39: Just confirming, this `silent:true` is only for the webpack logs, the nextjs ones  still go through?

package.json
```diff
@@ -104,4 +102,4 @@
     "vercel": "^21.2.3",
     "whatwg-fetch": "^3.5.0"
   }
-}
\ No newline at end of file
+}
```
solid-akb, line 105: is this a prettier thing? Is it possible to remove whitespace?

solid-akb (approved):
Thanks for finding this and implementing! Left a few small comments/questions but otherwise approve!

next.config.js
ThisIsMissEm: This was the configuration that was recommended by Sentry, also it only affects the sentry integration, as it's in `sentryWebpackPluginOptions`

package.json
ThisIsMissEm: it is a prettier thing, but in general, all files should end with a trailing whitespace.

PR #472, opened 2022-04-10
Chore: migrate from mock localstorage

ThisIsMissEm:
## Description

The dependency on `jest-mock-localstorage` was previously causing issues when upgrading Jest, and it turns out it's not actually needed. I also found that `jest-mock-extended` wasn't currently used anywhere.

This gives a stepping stone towards being able to upgrade jest, which would also allow for migrating the `@swc/jest` which would make running the tests significantly faster.

## Changes

See the individual commits

## Testing

Only includes changes to tests

## Commit checklist


solid-akb (approved):
Just left one comment. Great PR. Thanks!

PR #473, opened 2022-04-11
Clean up for Loading state and tests

solid-akb:
## Description
Was going to be to fix the bug but after deprioritizing still wanted to get some cleanup code in there. 
## Changes
There was some logic and nested ternaries that I think may have caused some unintended loading states. This PR cleans that up. Also there were a few tests that had the wrong props being sent in that have been fixed. 
## Testing
All tests pass.
## Commit checklist

## Screenshots/captures
Mini snippet of logic below: For example in line 159, if page was still loading it would go to the second return in the ternary even if there was no data. 
image

### DIFF
--- a/components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
+++ b/components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
@@ -253,7 +253,6 @@ export default function AgentAccessTable({ type, loading, setLoading }) {
         </>
       )}
       {tableContents()}
-      {/* <AgentAccessTableContent loading={loading} data={data} type={type} /> */}
     </Accordion>
   );
 }
--- a/components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
+++ b/components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
@@ -41,7 +41,6 @@ import AddAgentButton from "../addAgentButton";
 // import AgentsTableTabs from "../agentsTableTabs";
 import AgentsSearchBar from "../agentsSearchBar";
 import { POLICIES_TYPE_MAP } from "../../../../constants/policies";
-
 import styles from "./styles";
 import PolicyHeader from "../policyHeader";
 import PolicyActionButton from "../policyActionButton";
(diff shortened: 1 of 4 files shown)


components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
```diff
@@ -116,47 +115,28 @@ export default function AgentAccessTable({ type, loading, setLoading }) {
...
-            <AgentsSearchBar handleFilterChange={handleFilterChange} />
-          </>
-        )}
-        {loading && (
-          <Container variant="empty">
-            <CircularProgress />
-          </Container>
-        )}
-        {!loading && tablePermissions.length ? (
+  const tableContents = () => {
```
ThisIsMissEm, line 120: would be better to make this an explicit component, but keep it in the same file, that way you've props you can pass to it, and it won't be redeclared each render.

components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
```diff
@@ -191,53 +171,88 @@ export default function AgentAccessTable({ type, loading, setLoading }) {
...
+          {/* TODO: Uncomment to reintroduce tabs */}
+          {/* <Tabs */}
+          {/*  handleTabChange={handleTabChange} */}
+          {/*  selectedTabValue={selectedTabValue} */}
+          {/*  tabs={[{ label: "All", value: "all", testid: TESTCAFE_ID_TAB_ALL },  { label: "People", value: "agent", testid: TESTCAFE_ID_TAB_PEOPLE }, { label: "Groups", value: "group", testid: TESTCAFE_ID_TAB_GROUPS } }]} */}
+          {/* /> */}
+          <AgentsSearchBar handleFilterChange={handleFilterChange} />
+        </>
+      )}
+      {tableContents()}
```
ThisIsMissEm, line 255: Then here you'd have `<PermissionsTable ... />` with the various props it needs

ThisIsMissEm (approved):
Looks overall good, but one comment structure wise

components/resourceDetails/resourceSharing/agentAccessTable/index.jsx
solid-akb: Awesome - makes sense. Thanks!

PR #474, opened 2022-04-12
Copy resource link

solid-akb:
## Description
Now you can copy your resource address in PB. 
## Changes
New copy link button on Resource Details Drawer
## Testing
Navigate to a container or resource in your pod. Click on the open area next to the name. The details drawer should open to the right. Click on "COPY LINK" and your resource/container address will be saved to your clipboard. 
## Commit checklist

- [ x] All acceptance criteria are met.
- [ x] Includes tests to ensure functionality and prevent regressions.

## Interested parties
@VirginiaBalseiro @KyraAssaad @ThisIsMissEm 
## Screenshots/captures

https://user-images.githubusercontent.com/94387648/162857823-a684b1c9-a113-4dd3-9190-ecb906cc58db.mov

### DIFF
--- a/components/resourceDetails/index.jsx
+++ b/components/resourceDetails/index.jsx
@@ -140,16 +140,6 @@ export default function ResourceDetails({
               </DownloadLink>
             </ActionMenuItem>
             <ActionMenuItem>
-              <DeleteResourceButton
-                className={actionMenuBem("action-menu__trigger", "danger")}
-                resourceIri={datasetUrl}
-                name={displayName}
-                onDelete={onDelete}
-                onDeleteCurrentContainer={onDeleteCurrentContainer}
-                data-testid={TESTCAFE_ID_DELETE_BUTTON}
-              />
-            </ActionMenuItem>
-            <ActionMenuItem>
               <Tooltip
                 PopperProps={{
                   disablePortal: true,
@@ -168,6 +158,16 @@ export default function ResourceDetails({
                   COPY LINK
                 </Typography>
               </Tooltip>
+            </ActionMenuItem>
+            <ActionMenuItem>
+              <DeleteResourceButton
+                className={actionMenuBem("action-menu__trigger", "danger")}
+                resourceIri={datasetUrl}
+                name={displayName}
+                onDelete={onDelete}
+                onDeleteCurrentContainer={onDeleteCurrentContainer}
+                data-testid={TESTCAFE_ID_DELETE_BUTTON}
+              />
             </ActionMenuItem>
           </ActionMenu>
         </AccordionDetails>
--- a/components/resourceDetails/index.jsx
+++ b/components/resourceDetails/index.jsx
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React, { useContext } from "react";
+import React, { useContext, useState } from "react";
 import {
   Accordion,
   AccordionDetails,
@@ -28,6 +28,7 @@ import {
   Divider,
   List,
   ListItem,
+  Tooltip,
   Typography,
 } from "@material-ui/core";
 import { makeStyles } from "@material-ui/styles";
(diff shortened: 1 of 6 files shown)


components/resourceDetails/index.jsx
```diff
@@ -92,6 +95,13 @@ export default function ResourceDetails({
     getAccordionKey(dataset, "sharing"),
     false
   );
+  const handleResourceCopyClick = () => {
+    navigator.clipboard.writeText(datasetUrl);
```
ThisIsMissEm, line 99: You might wish to wrap this in a try/catch, though I'm sure this API is widely supported, I'm not sure on errors it could throw off top of head.

components/resourceDetails/index.test.jsx
```diff
@@ -70,6 +71,27 @@ describe("Resource details", () => {
...
+      <DatasetProvider solidDataset={dataset}>
+        <ResourceDetails />
+      </DatasetProvider>
+    );
+    const copyLink = screen.getByTestId(TESTCAFE_RESOURCE_ADDRESS_TOOLTIP);
+    userEvent.click(copyLink);
+    await waitFor(() => {
+      expect(copyLink).toBeInTheDocument();
+      expect(writeText).toHaveBeenCalled();
+    });
```
ThisIsMissEm, line 92: Generally multiple asserts in a waitFor is a bad practice, but until you can migrate to userEvebts from testing-lib you're kinda stuck with events being dispatched async from test execution.

With userEvent, you can just do: 

```
await user.click(copyLink);

expect(...)
exoect(...)
```

ThisIsMissEm (approved):
Looks good to me! Nice one!

VirginiaBalseiro (approved):
looks great! 🚀

KyraAssaad (commented):
Great work! One small thing: could we make the button font-weight Medium instead of ExtraBold? I didn't see it in your screenshots but it looks like this on my machine: 
Inrupt_PodBrowser

solid-akb:
> Great work! One small thing: could we make the button font-weight Medium instead of ExtraBold? I didn't see it in your screenshots but it looks like this on my machine: Inrupt_PodBrowser

oh weird! What browser are you using? That should def be changed!

solid-akb:
@KyraAssaad I tested on Safari 15.4, Chrome 100.0.4896 and Firefox 100.0b4 and all look ok. Also it looks like the styling (which I copied from the other action items) is controlled in Prism 😖. Let's take a look tomorrow to see what version browser you are running and determine best path forward.

KyraAssaad:
@solid-akb so weird! I am on Chrome v100.0.4896.88 and I tried on Safari v15.4 and I still get the same. I wonder if it is somehow related to the fact that I have Raleway installed on my computer?

It looks like the class `MuiTypography-body1` is responsible for overriding the `Raleway-Medium` weight with `Raleway-ExtraBold`:
image

If it's because this is all controlled by Prism and will be a pain to change, then I'm fine leaving it the way it is.

components/resourceDetails/index.jsx
```diff
@@ -129,6 +140,26 @@ export default function ResourceDetails({
...
+                  disablePortal: true,
+                }}
+                open={tooltipOpen}
+                disableFocusListener
+                disableHoverListener
+                disableTouchListener
+                title="Copied"
+                data-testid={TESTCAFE_RESOURCE_ADDRESS_TOOLTIP}
+              >
+                <Typography
```
ThisIsMissEm, line 155: This should actually be a button, not just text:

See the `Delete Button` as an example: https://github.com/inrupt/pod-browser/blob/2fcfe4583c8092f48bb495c6ef9f0c2746c84a84/components/deleteButton/index.jsx#L120-L126

Or the Download button, which is the simpler one and more preferred: https://github.com/inrupt/pod-browser/blob/914d0406cea0a9fa699886b1a20f829fc4a4094c/components/downloadLink/index.jsx#L62

ThisIsMissEm (changes requested):

components/resourceDetails/index.jsx
solid-akb: Done!

ThisIsMissEm (approved):

KyraAssaad (approved):

PR #475, opened 2022-04-13
Feat: use access grants method to get request from IRI instead of full VC

VirginiaBalseiro:
## Description

This PR uses the new function in the access grants lib to fetch the access request. Also some cleanup, a small bug fix, and removing the ConsentRequestContext which is no longer needed.

## Changes

- fetch access request using getAccessRequestFromRedirectUrl

- remove ConsentRequestContext and related code since no longer necessary

- fix bug with confirmation dialog confirming instead of canceling when no purposes selected

## Testing

- generate a request using the demo app and check that everything is ok (appending to the Vercel deployment for this PR's URL)

## Commit checklist


components/consentRequestForm/index.jsx
```diff
@@ -206,6 +213,14 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
     setSelectedDate(date ? date.toISOString() : null);
   };
 
+  useEffect(() => {
+    if (!consentRequest) return;
+    setRequestedAccesses(getRequestedAccesses(consentRequest));
+    setExpirationDate(getExpiryDate(consentRequest));
+    setPurposes(getPurposeUrls(consentRequest));
+    setResourceOwnerWebId(getDataSubjectWebId(consentRequest));
+  }, [consentRequest]);
```
ThisIsMissEm, line 222: This doesn't need to be a `useEffect` I don't think — I think you can just use initial state:

```
  const [requestedAccesses, setRequestedAccesses] = useState(() => getRequestedAccesses(consentRequest));
  const [expirationDate, setExpirationDate] = useState(() => getExpiryDate(consentRequest));
  const [purposes, setPurposes] = useState(() => getPurposeUrls(consentRequest));
  const [resourceOwnerWebId, setResourceOwnerWebId] = useState(() => getDataSubjectWebId(consentRequest));
```

And just have those methods return an empty value or null if there isn't a `consentRequest` (though there always should be one, as the prop is marked as required)

ThisIsMissEm: That is, if the `consentRequest` changes the entire component will be re-rendered from scratch anyway, as the prop changed.

components/consentRequestForm/index.jsx
```diff
@@ -325,6 +345,7 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
 }
 
 ConsentRequestForm.propTypes = {
+  consentRequest: T.objectOf(T.any).isRequired,
```
ThisIsMissEm, line 348: It'd probably be ideal to use a proper `T.shape` here instead of just `T.object(T.any)`, though you could definitely make that shape just include the properties you use, I think.

components/pages/privacy/access/requests/index.jsx
```diff
@@ -39,35 +43,36 @@ import {
...
 
   useEffect(() => {
     if (!consentRequest) return;
     (async () => {
-      const { dataset: profileDataset } = await getProfileResource(
-        agentWebId,
-        fetch
-      );
+      const profiles = await getProfileAll(agentWebId);
+      const profileDataset = profiles.webIdProfile || profiles.altProfileAll[0]; // getting the first profile for now
```
ThisIsMissEm, line 75: you could also do just `profiles.altProfileAll.first((profile) => !!profile)`, that is "find the first alternative profile which could be loaded"

components/pages/privacy/access/requests/index.jsx
```diff
@@ -39,35 +43,36 @@ import {
...
 import Spinner from "../../../../spinner";
 
 const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
 export default function ConsentShow() {
   const { session } = useSession();
   const { fetch } = session;
-  const router = useRouter();
-  const { requestVc } = router.query;
+  const redirectUrl = window.location.href;
```
ThisIsMissEm, line 53: I probably wouldn't store this as a constant, just do it inline in the `useEffect`, only `session.fetch` may change, but even then, that's unlikely

components/pages/privacy/access/requests/index.jsx
```diff
@@ -39,35 +43,36 @@ import {
...
   const bem = useBem(useStyles());
   const [consentRequest, setConsentRequest] = useState(null);
   const agentWebId = getRequestorWebId(consentRequest);
   const [agentDetails, setAgentDetails] = useState({});
   const { agentName, agentUrl, agentPolicy, agentTOS } = agentDetails || null;
```
ThisIsMissEm, line 58: An alternative to having all these states would be to lift the loading logic up into a function that returns back data in the shape you need it, then you'd just have:

```
const [isLoading, setLoading] = useState(true); // default to loading
const [error, setError] = useState(null);
const [requestDetails, setRequestDetails] = useState(null);

useEffect(() => {
  // You could also wrap this in a `Promise.race()` with a `sleep` to provide a timeout error.
  fetchDetails(window.location.href).then(
    (details) => setRequestDetails(details),
    (err) => setError(err)
  ).finally(() => setLoading(false));
});
```

And then have the `fetchDetails` method handle loading everything required, returning back an object of `{ consentRequest, agentWebId, agentDetails }`, as I don't think any of the individual state would change, as it's all predicated on the `window.location.href`

This would reduce the amount of back-and-forth react has to do managing the state, and make your code more testable.

components/pages/privacy/access/requests/index.jsx
```diff
@@ -87,73 +92,66 @@ export default function ConsentShow() {
...
-                </span>
-              )}
+    <MuiPickersUtilsProvider utils={DateFnsUtils}>
+      <Container className={bem("request-container")}>
+        <ConsentRequestForm
+          consentRequest={consentRequest}
+          agentDetails={agentDetails}
+          agentWebId={agentWebId}
+        />
+        <div className={bem("request-container__content")}>
```
ThisIsMissEm, line 102: It might be nice to move this to an `AgentDetails` component that takes in `agentDetails` and `agentWebId`, though not strictly necessary (but I think this UI might reoccur through out the application)

ThisIsMissEm (commented):
Left a few comments about how one might improve this code, but it does do it's job for now.

components/consentRequestForm/index.jsx
```diff
@@ -213,14 +213,6 @@ export default function ConsentRequestForm({
...
 
-  useEffect(() => {
-    if (!consentRequest) return;
-    setRequestedAccesses(getRequestedAccesses(consentRequest));
-    setExpirationDate(getExpiryDate(consentRequest));
-    setPurposes(getPurposeUrls(consentRequest));
-    setResourceOwnerWebId(getDataSubjectWebId(consentRequest));
-  }, [consentRequest]);
-
   useEffect(() => {
```
ThisIsMissEm, line 216: `purposes` never changes here

components/consentRequestForm/index.jsx
```diff
@@ -325,6 +330,21 @@ export default function ConsentRequestForm({ agentDetails, agentWebId }) {
 }
 
 ConsentRequestForm.propTypes = {
+  consentRequest: T.shape({
```
solid-akb, line 333: Nice!

components/confirmationDialogWithProps/index.jsx
```diff
@@ -99,18 +100,20 @@ export default function ConfirmationDialogWithProps({
...
-            isDangerousAction
-              ? classes.dangerButton
-              : classes.submitAgentsButton
-          }
-          type="submit"
-          onClick={onConfirm}
-        >
-          {confirmText || "Confirm"}
-        </Button>
+        {!omitConfirmButton && (
```
solid-akb, line 103: Thanks for doing this in two components

components/consentRequestForm/index.jsx
```diff
@@ -71,20 +70,38 @@ export const CONFIRM_TEXT = "Continue with no access";
...
+    }
+    if (!Array.isArray(purposes)) {
+      return purposes;
+    }
+    return [];
+  });
   const selectedResources = selectedAccess.map(
     ({ resourceIri }) => resourceIri
   );
+  // other state
```
solid-akb, line 104: why "local state" and "other state"?

solid-akb (approved):

components/consentRequestForm/index.jsx
VirginiaBalseiro: I mean related to the request vs not

PR #476, opened 2022-04-13
fixed PropType errors

solid-akb:
## Description
Fixed prop error on Survey widget
## Commit checklist

### DIFF
--- a/components/surveyWidget/widgetBox/index.jsx
+++ b/components/surveyWidget/widgetBox/index.jsx
@@ -84,7 +84,7 @@ WidgetBox.propTypes = {
   id: T.string,
   open: T.bool,
   // eslint-disable-next-line react/forbid-prop-types
-  anchorEl: T.any,
+  anchorEl: T.object,
   onClose: T.func.isRequired,
 };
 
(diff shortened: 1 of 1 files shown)


components/surveyWidget/widgetBox/index.jsx
```diff
@@ -82,7 +83,8 @@ export default function WidgetBox({ id, open, anchorEl, onClose }) {
 WidgetBox.propTypes = {
   id: T.string,
   open: T.bool,
-  anchorEl: T.element,
+  // eslint-disable-next-line react/forbid-prop-types
+  anchorEl: T.any,
```
ThisIsMissEm, line 87: Why wouldn't this be `T.element`? Or should it have actually be a `T.node` instead? See https://dev.to/fromaline/jsxelement-vs-reactelement-vs-reactnode-2mh2 for the difference being kind of explained

solid-akb: Neither T.element or T.node work. I updated to say T.object which is more explicit but it still requires an es lint exception

ThisIsMissEm: Strange! Those _should_ work if anchorEl is a jsx element.. but whatever, let's just prevent the warning & consider it debt to avoid in the future

ThisIsMissEm: Strange! Those _should_ work if anchorEl is a jsx element.. but whatever, let's just prevent the warning & consider it debt to avoid in the future

PR #478, opened 2022-04-15
Default to WebID on profile page if the user has not provided a name

solid-akb:
## Description
Default to WebID on profile page if the user has not provided a name
## Changes
If the user does not add a Name to their profile (there is no name by default) we get an error on the Profile Page. Now we show their name if they provided a name or a WebID if they have not provided a name. Also removed the link because it is now unnecessary.
## Testing
See video below
## Commit checklist

- [ x] All acceptance criteria are met.
- [x ] Includes tests to ensure functionality and prevent regressions.
- [x ] The changelog has been updated, if applicable.

## Interested parties
@VirginiaBalseiro @KyraAssaad @ThisIsMissEm 
## Screenshots/captures

https://user-images.githubusercontent.com/94387648/163581018-0c4a48c1-b65e-4bae-b108-c1acaf21ec2c.mov


ThisIsMissEm:
@KyraAssaad how do you feel about the profile page having no save button or loading indicators to show that something's happening/happened?

Just noticed it looking weird when watching the screen recording above here

components/profile/personAvatar/index.test.jsx
```diff
@@ -61,6 +62,35 @@ describe("Person Avatar", () => {
...
+    );
+
+    const session = mockSession();
+    const SessionProvider = mockSessionContextProvider(session);
+    const { getByText } = renderWithTheme(
+      <CombinedDataProvider
+        solidDataset={profileDataset}
+        thing={profileThingWithNameRemoved}
+      >
+        <SessionProvider>
```
ThisIsMissEm, line 84: SessionProvider should be on the outside of the dataset provider

components/profile/personAvatar/index.test.jsx
```diff
@@ -61,6 +62,35 @@ describe("Person Avatar", () => {
...
+          <PersonAvatar profileIri={profileIri} />
+        </SessionProvider>
+      </CombinedDataProvider>
+    );
+
+    await waitFor(() => {
+      expect(getByText(session.info.webId)).toBeInTheDocument();
```
ThisIsMissEm, line 91: Would be better to assert that the text content of the header with the test id equals the webid, incase something else adds the webid to the page

ThisIsMissEm (approved):

components/profile/personAvatar/index.test.jsx
solid-akb: I like that - will update

components/profile/personAvatar/index.test.jsx
solid-akb: fixed

components/profile/personAvatar/index.test.jsx
solid-akb: updated

PR #480, opened 2022-04-21
Refactor: profiles

VirginiaBalseiro:
This PR aims to solve a number of issues and clear assumptions we have made in our handling of Profiles.

It introduces a new function that follows [discovery steps to get a full profile](https://github.com/solid/webid-profile/blob/main/notes/pre-final-draft.md#2-discovering-a-complete-solid-profile) and return as an object. Ideally we would be able to combine datasets via the SDK components (dataset providers?), but in this way we get all the profile data we need for PB to display profiles in the Privacy page, Contacts, etc. 

For the user's own Profile page, which should be editable, we use whatever editable profile document we can find first. 

Note that for all of these pieces of profile information we default to the first item, this is until we have decision with regards to how to handle multiple profiles on PB. 

There is certainly a lot more work to be done and **there are certain things that this PR does not cover**, like:

- Refactoring contacts so that we don't fetch profile datasets anymore, which would involve moving away from the SDK Table component.
- Remove all dependency on old ways of fetching profiles, like the `fetchProfile` function, and update all hooks that fetch users/agents profiles.
- Consolidating the multiple ways we display profiles: just Avatars when no profile available (fallback and WebID), Avatar and Profile in Privacy page with Tabs, entire Profile in Contacts, entire editable Profile in Profile page... and so on. 

All of these should be tackled, but depending on where we want to go with these features (Contacts, etc.), it might have been unnecessary work at the moment. This PR addresses the immediate issues (in particular, a regression bug where the wrong ID is displayed on all profiles), while still trying to provide a full profile that our components can make use of, as well as an example of how to follow the discovery steps to get the full profile. 

## Testing

- To test, check that everything works as intended in Profiles on:
  - Privacy page
  - Contacts
  - User's profile page (editable)

### DIFF
--- a/components/pages/privacy/resourceAccess/show/index.jsx
+++ b/components/pages/privacy/resourceAccess/show/index.jsx
@@ -22,10 +22,10 @@
 /* eslint-disable react/jsx-props-no-spreading */
 /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 import React, { useEffect, useState, useMemo } from "react";
-import { CombinedDataProvider, useSession, Text } from "@inrupt/solid-ui-react";
+import { useSession } from "@inrupt/solid-ui-react";
 import { useRouter } from "next/router";
 import T from "prop-types";
-import { schema, foaf } from "rdf-namespaces";
+import { schema } from "rdf-namespaces";
 import { makeStyles } from "@material-ui/styles";
 import { useSortBy, useTable } from "react-table";
 import clsx from "clsx";
@@ -39,7 +39,7 @@ import {
   Icons,
   Table as PrismTable,
 } from "@inrupt/prism-react-components";
-import { createStyles, Divider } from "@material-ui/core";
+import { createStyles, Divider, Typography } from "@material-ui/core";
 import ConfirmationDialog from "../../../../confirmationDialog";
 import { handleAction } from "../../../../containerTableRow";
 import PersonAvatar from "../../../../profile/personAvatar";
@@ -58,6 +58,7 @@ import { isContainerIri } from "../../../../../src/solidClientHelpers/utils";
 import { getParentContainerUrl } from "../../../../../src/stringHelpers";
 import TabPanel from "./tabPanel";
 import styles from "./styles";
+import useFullProfile from "../../../../../src/hooks/useFullProfile";
 
 const useStyles = makeStyles((theme) => createStyles(styles(theme)));
 
@@ -69,7 +70,6 @@ export default function AgentResourceAccessShowPage({ type }) {
   const router = useRouter();
   const decodedIri = decodeURIComponent(router.query.webId);
   const tableClass = PrismTable.useTableClass("table", "inherits");
-
   const { session, fetch } = useSession();
   const podRoot = usePodRootUri(session.info.webId);
   const [resources, setResources] = useState([]);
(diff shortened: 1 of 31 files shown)


components/pages/privacy/resourceAccess/show/index.jsx
```diff
@@ -216,199 +219,181 @@ export default function AgentResourceAccessShowPage({ type }) {
...
+                        {headerGroup.headers.map((column) => (
+                          <td
+                            key={column.id}
+                            className={bem("table__head-cell")}
+                            {...column.getHeaderProps(
+                              column.getSortByToggleProps()
+                            )}
+                          >
+                            {column.render("Header")}
+                            {` `}
```
solid-akb, line 277: do we want this here or typo?

components/pages/privacy/resourceAccess/show/index.jsx
```diff
@@ -216,199 +219,181 @@ export default function AgentResourceAccessShowPage({ type }) {
...
     return <Spinner />;
   }
-  // FIXME: using a conditional dataset provider wrapper while we mock app profiles
-  const ConditionalWrapper = ({ condition, wrapper, children }) =>
-    condition ? wrapper(children) : children;
 
   const renderAvatar = () => {
-    if (type === schema.SoftwareApplication) {
-      return <AppAvatar profileIri={decodedIri} />;
+    if (agentTypes.includes(schema.SoftwareApplication)) {
```
solid-akb, line 224: is there anyway we can export this function somewhere global and use it whenever we render an avatar (and then get rid of the other avatar components?)

solid-akb (commented):
This is an awesome PR! Nice work! I left a few comments - the biggest being moving the avatar function into an Avatar component that would replace the other ~6 that we have but that's obviously a bigger lift so if you think that should be in it's own PR I am happy with that.

components/pages/privacy/resourceAccess/show/index.jsx
VirginiaBalseiro: Let's keep this as in for now. As mentioned in the PR, description, we have different "versions" of Avatars and need to do more work on it but I would make that a separate PR: 

> Consolidating the multiple ways we display profiles: just Avatars when no profile available (fallback and WebID), Avatar and Profile in Privacy page with Tabs, entire Profile in Contacts, entire editable Profile in Profile page... and so on.

components/pages/privacy/resourceAccess/show/index.jsx
```diff
@@ -216,199 +219,181 @@ export default function AgentResourceAccessShowPage({ type }) {
...
+        {renderAvatar()}
+        <Tabs
+          tabs={tabs}
+          handleTabChange={handleTabChange}
+          selectedTabValue={selectedTabValue}
+        />
+        <TabPanel value={selectedTabValue} panelName="Permissions">
+          <p>
+            {agentProfile ? agentProfile.names[0] : decodedIri}
+            {` `}
```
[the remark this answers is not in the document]
VirginiaBalseiro: yes

components/pages/privacy/resourceAccess/show/index.jsx
VirginiaBalseiro: we do need it

solid-akb (approved):

PR #482, opened 2022-05-04
Fix/update contacts

VirginiaBalseiro:
This PR addresses a bug where 2.0 account could not be added to contacts, as the old `fetchProfile` function failed to get a profile document. In this PR we use the new `getFullProfile` function to get a full Solid profile and use that instead in the specific places where necessary to fix this bug.

This PR ***does not*** intend to be a deep cleanup of Contacts code, it just addresses the bug, as updating the code in Contacts completely would be a much larger undertaking and it is not worth addressing now considering other priorities.

Essentially this PR updates: 
- AddContact button to use new full profile function
- ContactList (specifically, ProfileLink) to use name from profile returned from new full profile function

## Testing
- try to add a 2.0 WebID to contacts and verify it works as intended

## Commit checklist

### DIFF
--- a/components/addContact/index.jsx
+++ b/components/addContact/index.jsx
@@ -33,7 +33,10 @@ import AgentSearchForm from "../agentSearchForm";
 import DetailsMenuContext from "../../src/contexts/detailsMenuContext";
 import AlertContext from "../../src/contexts/alertContext";
 import styles from "./styles";
-import { fetchProfile } from "../../src/solidClientHelpers/profile";
+import {
+  fetchProfile,
+  getFullProfile,
+} from "../../src/solidClientHelpers/profile";
 import { isPodOwner } from "../../src/solidClientHelpers/utils";
 import useContactsOld from "../../src/hooks/useContactsOld";
 import useContactsContainerUrl from "../../src/hooks/useContactsContainerUrl";
@@ -67,7 +70,7 @@ export function handleSubmit({
     setIsLoading(true);
 
     try {
-      const { name, webId, types } = await fetchProfile(iri, fetch);
+      const { names, webId, types } = await getFullProfile(iri, session);
       const existingContact = await findContactInAddressBook(
         people,
         webId,
@@ -80,7 +83,7 @@ export function handleSubmit({
         return;
       }
 
-      const contact = { webId, fn: name || null };
+      const contact = { webId, fn: names[0] || null };
       const { response, error } = await saveContact(
         addressBook,
         addressBookContainerUrl,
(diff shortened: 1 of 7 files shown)


components/addContact/index.jsx
```diff
@@ -67,28 +70,28 @@ export function handleSubmit({
...
       if (existingContact) {
         alertError(EXISTING_WEBID_ERROR_MESSAGE);
         setIsLoading(false);
         return;
       }
-
-      const contact = { webId, fn: name || null };
+      const contact = {
+        webId,
+        fn: names[0],
```
ThisIsMissEm, line 86: I assume `names[0]` is correct and that we wouldn't want to do a `first`/`find` operation to select the first name that has a value?

VirginiaBalseiro: no null/undefined values are put into that array

components/profileLink/index.jsx
```diff
@@ -37,34 +38,19 @@ export function buildProfileLink(webId, route, path) {
...
   const { route } = useRouter();
-  const { webId } = props;
   const { thing } = useThing();
-  const contact = getProfileIriFromContactThing(thing);
-
-  const type = getUrl(thing, rdf.type);
-  const path = type === schema.Person ? "person" : "app";
+  const webId = getProfileIriFromContactThing(thing);
+  const profile = useFullProfile(webId);
+  const path = profile.types.includes(schema.Person) ? "person" : "app";
```
ThisIsMissEm, line 46: On the first render of the `useFullProfile` hook, it returns `null`, as such you'd need to check if you've a `profile` value before looking at `profile.types.includes`

So you'll probably want a `if (profile === null) return null;` in here.

The comments below should probably also be removed:

components/profileLink/index.test.jsx
```diff
@@ -48,7 +51,12 @@ describe("ProfileLink", () => {
       route: "/contacts",
     });
   });
-  it("supports rendering name from vcard.fn", () => {
+  it("renders first name from profile names array", () => {
```
ThisIsMissEm, line 54: make sure you've a test for `mockedUseFullProfile.mockReturnValue(null)` as that's what the hook initially returns on first render.

ThisIsMissEm (changes requested):
Just need to fix the usage of useFullProfile to account for it returning null on first render

components/addContact/index.jsx
ThisIsMissEm: okay! I was thinking more if there was a profile you could read or the first profile didn't have a name or some such. Though I'm not super familiar with the code, hence a comment asking :)

solid-akb (commented):
Looks good! Saw Emelia left some comments to address, otherwise good for merge

ThisIsMissEm (approved):

PR #483, opened 2022-05-16
Fix: refresh panel after revoking access grant

VirginiaBalseiro:
This PR includes adding a trigger to refresh access grants permissions after revoking, but also some code cleanup with involved updating more use cases of profiles in some of the components involved.

- trigger refresh of access grants permissions when revoking

- simplify code by removing usePermissionsWithProfiles hook

- update the way profiles are fetched and displayed for individual agent permission/profile in the sharing panel

## Testing

- Test that revoking an access grant from the sharing panel triggers a refresh and the agent is removed from the list.
- Test that permissions work properly. I tested with NSS, ESS 1.1 and ESS 2.0.

## Commit checklist


solid-akb:
Hey! Great PR and thanks for doing some other clean up with it! 
Great news is that the panel did refresh automatically when I changed access on a resource. 
Unfortunately, I also saw some strange behavior as well.  See video below: there is like a ghost permission that is in a constant loading state. When I revoked the legitimate permission, the ghost permission was still there.

https://user-images.githubusercontent.com/94387648/168598283-d49158a4-cebe-4cc2-aef7-198f4bd847a8.mov

solid-akb:
Some other thoughts that won't block my approval but things we should look at when we do PB 2.0. 

We shouldn't be triggering the reload with "shouldUpdate" states that we manually manipulate. The hooks should automatically refresh the data when it changes. The shouldUpdate is a workaround for now, but let's try and avoid that in the future. 

Instead of `rdf-namespaces` library we should start using `import { SCHEMA_INRUPT } from "@inrupt/vocab-common-rdf"`

There is also a neat fetch that gets passed in with the auth library. This avoids us having to import session and then de-structure fetch from it. `import { fetch } from "@inrupt/solid-client-authn-browser"`

VirginiaBalseiro:
> Hey! Great PR and thanks for doing some other clean up with it! Great news is that the panel did refresh automatically when I changed access on a resource. Unfortunately, I also saw some strange behavior as well. See video below: there is like a ghost permission that is in a constant loading state. When I revoked the legitimate permission, the ghost permission was still there.
> 
>  loading.mov

@solid-akb can you test again please?

solid-akb:
Looks good!

solid-akb (approved):

PR #485, opened 2022-06-02
Fix/rdf upload

VirginiaBalseiro:
In this PR:

- using contentType instead of type when overwriting file
- using text/turtle if the file has a ttl extension and no type after upload
- updating existing tests and add test for turtle files


solid-akb (approved):
Looks good!

PR #488, opened 2022-06-06
Update VC query param for access grants

ajacksified:
This updates the hardcoded VC query parameter used in the access grant redirect flow to use a variable exported by the access-grants library.


components/pages/privacy/access/requests/generate-access-request/index.jsx
```diff
@@ -59,7 +59,7 @@ export default function GenerateAccessRequest() {
     if (vc && origin) {
       redirectToAccessManagementUi(
         vc,
-        `${origin}/privacy/?signedVcUrl=${encodeURIComponent(vc.id)}`,
+        `${origin}/privacy/?accessGrantUrl=${encodeURIComponent(vc.id)}`,
```
matthieubosquet, line 62: So I guess it has to be hardcoded here?

PR #491, opened 2022-06-09
set UTC hours when passing to approve fn instead of when selecting date

VirginiaBalseiro:
## Description

Setting UTC hours before passing to approve function instead of when setting date on picker.

### DIFF
--- a/components/accessRequestForm/index.jsx
+++ b/components/accessRequestForm/index.jsx
@@ -127,6 +127,8 @@ export default function AccessRequestForm({
   const handleSubmit = async (e) => {
     e.preventDefault();
     if (selectedAccess.length && selectedPurposes.length) {
+      const date = new Date(selectedDate);
+      date.setUTCHours(23, 59, 59, 0);
       const signedVc = await approveAccessRequest(
         session.info.webId,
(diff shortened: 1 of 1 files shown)


matthieubosquet (approved):

components/accessRequestForm/index.jsx
```diff
@@ -127,14 +127,16 @@ export default function AccessRequestForm({
   const handleSubmit = async (e) => {
     e.preventDefault();
     if (selectedAccess.length && selectedPurposes.length) {
+      const date = new Date(selectedDate);
+      date.setUTCHours(23, 59, 59, 0);
```
matthieubosquet, line 131: @VirginiaBalseiro I just realised one small extra but, selectedDate is not necessarily set here.

So if we want to be able to do the "never expiring" VCs, we need:

```javascript
let expirationDate;
if (selectedDate) {
    expirationDate = new Date(selectedDate);
    expirationDate.setUTCHours(23, 59, 59, 0);
}
```

And below we can remove assignment:
```javascript
          resources: selectedResources,
          expirationDate,
          resourceOwner: session.info.webId,
```

PR #495, opened 2022-07-15
add new thing with correct subject if not available

VirginiaBalseiro:
## Description

This PR: 
- introduces a change in the `getFullProfile` function: when a Thing with WebID as subject cannot be found, one is created and added to the dataset (all of the editable datasets) so that we can use the WebID as Thing URL in the Context wrapping the Profile page.
- simplifies the `getFullProfile` function
- add tests

## To test:
- Create a new account and edit profile on prod PB
- Verify the profile contains wrong triples (profile dataset URL as subject instead of WebID)
- Log in on the deployment of this branch and edit profile
- verify a triple has been added with WebID as subject

## Commit checklist

### DIFF
--- a/components/header/userMenu/index.jsx
+++ b/components/header/userMenu/index.jsx
@@ -40,14 +40,13 @@ export default function UserMenu() {
   const bem = useBem(useStyles());
   const authenticatedProfile = useAuthenticatedProfile();
   const menu = useUserMenu();
-
   if (!authenticatedProfile) return <Spinner />;
 
   return (
     <div className={bem("userMenu")} data-testid={TESTCAFE_ID_USER_MENU}>
       <PrismUserMenu
         className={bem("userMenu__trigger")}
-        label={authenticatedProfile.name || "Account"}
+        label={authenticatedProfile.names[0] || "Account"}
         imgSrc={authenticatedProfile.avatar}
         menu={menu}
         menuId="UserMenu"
--- a/components/profile/personProfile/index.jsx
+++ b/components/profile/personProfile/index.jsx
@@ -21,7 +21,7 @@
 
 import React from "react";
 import T from "prop-types";
-import { getSourceUrl } from "@inrupt/solid-client";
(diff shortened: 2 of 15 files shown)


solid-akb:
I ran into unexpected behavior. I verified that Prod is updating the Profile doc incorrectly (with the storage instead of WebID). When I went to update my profile in the local branch this is what the profile page looks like. None of the fields are editable. Was using the User I just created in PB prod to log in to the local PB. 

image

components/profile/personProfile/index.jsx
```diff
@@ -21,7 +21,7 @@
 
 import React from "react";
 import T from "prop-types";
-import { getSourceUrl } from "@inrupt/solid-client";
+import { getSourceUrl, getThing } from "@inrupt/solid-client";
```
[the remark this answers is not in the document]
VirginiaBalseiro: oops - I was trying something and forgot to remove

VirginiaBalseiro:
@solid-akb could you test again? Thanks!!

components/header/userMenu/index.jsx
```diff
@@ -40,14 +40,13 @@ export default function UserMenu() {
...
   const menu = useUserMenu();
-
   if (!authenticatedProfile) return <Spinner />;
 
   return (
     <div className={bem("userMenu")} data-testid={TESTCAFE_ID_USER_MENU}>
       <PrismUserMenu
         className={bem("userMenu__trigger")}
-        label={authenticatedProfile.name || "Account"}
+        label={authenticatedProfile.names[0] || "Account"}
```
chelseapinka, line 49: Just for for my better understanding, is it ever possible for this array of names to have more than one value?

VirginiaBalseiro: Yes

solid-akb:
Profile updating looks good!

solid-akb (approved):

PR #500, opened 2022-07-20
temporarily block login on podspaces, display warning

VirginiaBalseiro:
## Description

Temporarily block login on Podspaces until migration is completed

## Commit checklist

### DIFF
--- a/components/login/index.jsx
+++ b/components/login/index.jsx
@@ -92,9 +92,9 @@ export default function LoginForm() {
       {shouldDisplayWarning && (
         <Message variant="warning" prominent hasIcon>
           Sorry, we cannot log you in at the moment due to planned system
-          maintenance. Please see our
-          <a href="https://inrupt.com/blog/pod-spaces-upgrade"> blog post </a>
-          for more information.
+          maintenance. Please see our{" "}
+          <a href="https://inrupt.com/blog/pod-spaces-upgrade">blog post</a> for
+          more information.
         </Message>
       )}
       <h3 data-testid={TESTCAFE_ID_LOGIN_TITLE}>Sign in with</h3>
--- a/components/login/index.jsx
+++ b/components/login/index.jsx
@@ -23,7 +23,7 @@
 
 import React, { useCallback, useState, useEffect } from "react";
 import { createStyles, makeStyles } from "@material-ui/styles";
-import { Button } from "@inrupt/prism-react-components";
+import { Button, Message } from "@inrupt/prism-react-components";
 import { useSession } from "@inrupt/solid-ui-react";
 import { useBem } from "@solid/lit-prism-patterns";
 import InfoTooltip from "../infoTooltip";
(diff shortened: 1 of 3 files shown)


chelseapinka (approved):
Makes sense and looks nice to have message appear on Signin button click.

components/login/index.jsx
```diff
@@ -45,12 +45,21 @@ const useStyles = makeStyles((theme) => createStyles(styles(theme)));
...
 
   const login = useCallback(
     (oidcIssuer) => {
+      if (
+        oidcIssuer.includes("login.inrupt.com") ||
+        oidcIssuer.includes("broker.pod.inrupt.com")
+      ) {
+        setShouldDisplayWarning(true);
+        return;
+      }
```
ThisIsMissEm, line 62:
```suggestion
      } else {
         setShouldDisplayWarning(false);
      }
```

components/login/index.test.jsx
```diff
@@ -118,4 +119,27 @@ describe("Login form", () => {
...
+
+    act(() => {
+      signinButton.click();
+    });
+
+    expect(mockUseReturnUrl.persist).not.toHaveBeenCalled();
+    expect(session.login).not.toHaveBeenCalled();
+
+    expect(asFragment).toMatchSnapshot();
+  });
```
ThisIsMissEm, line 144: Maybe add a test case proving that CSS/NSS could be used? idk.

components/login/index.jsx
VirginiaBalseiro: If `shouldDisplayWarning` is set to false by default, and only changed to true if this condition is true, do we need this?

ThisIsMissEm: Well, you can type in one idp, then get the warning, then type in a different one

VirginiaBalseiro: but if that happens you're redirected, as this is the login function and it is not called on value change, so the warning will stay there even when typing and won't clear until it is submitted

components/login/index.test.jsx
VirginiaBalseiro: Added test - passed an IDP value via query params to save myself the whole clicking through the dropdown thing.

ThisIsMissEm (approved):
Looks good to me.

ThisIsMissEm:
wait.

components/login/index.jsx
```diff
@@ -79,6 +88,15 @@ export default function LoginForm() {
 
   return (
     <div className={bem("login-form")}>
+      {/* FIXME: remove after pod migration is completed */}
+      {shouldDisplayWarning && (
+        <Message variant="warning" prominent hasIcon>
+          Sorry, we cannot log you in at the moment due to planned system
+          maintenance. Please see our
+          <a href="https://inrupt.com/blog/pod-spaces-upgrade"> blog post </a>
+          for more information.
```
ThisIsMissEm, line 97:
```suggestion
          maintenance. Please see our{" "}
          <a href="https://inrupt.com/blog/pod-spaces-upgrade">blog post</a>{" "}
          for more information.
```

To fix the link underline:
Screenshot 2022-07-20 at 17 22 52

ThisIsMissEm (changes requested):

ThisIsMissEm (approved):

PR #502, opened 2022-07-21
use new profile function in auth profile hook

VirginiaBalseiro:
## Description

The `useAuthenticatedHook` was using an old fetch profile function that defaults to an extended profile if available, ignoring the WebID profile, and therefore missing the storage, which is why PB was hanging because no pod was available.

### DIFF
--- a/components/container/index.jsx
+++ b/components/container/index.jsx
@@ -77,8 +77,7 @@ function maybeRenderWarning(locationIsInUsersPod, noControlError, podRootIri) {
 export default function Container({ iri }) {
   const { sessionRequestInProgress } = useSession();
   const [resourceUrls, setResourceUrls] = useState(null);
-  const { data: authenticatedProfile, error: authenticatedProfileError } =
-    useAuthenticatedProfile();
+  const authenticatedProfile = useAuthenticatedProfile();
   const podRootIri = usePodRootUri(iri);
   const [noControlError, setNoControlError] = useState(null);
   const { data: podRootResourceInfo, error: podRootError } =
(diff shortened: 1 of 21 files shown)


components/container/index.jsx
```diff
@@ -77,8 +77,7 @@ function maybeRenderWarning(locationIsInUsersPod, noControlError, podRootIri) {
 export default function Container({ iri }) {
   const { sessionRequestInProgress } = useSession();
   const [resourceUrls, setResourceUrls] = useState(null);
-  const { data: authenticatedProfile, error: authenticatedProfileError } =
-    useAuthenticatedProfile();
+  const authenticatedProfile = useAuthenticatedProfile();
```
chelseapinka, line 80: is there something wrong with using the authenticatedProfileError? is that why its getting removed or is not getting the right error types?

VirginiaBalseiro: The reason is that `useFullProfile` doesn't return an error and just leaves the profile as null if something fails, otherwise it will return whatever values it can fetch in the profile object. I could change this to return an error if necessary.

components/pages/privacy/resourceAccess/show/index.jsx
```diff
@@ -117,15 +119,13 @@ export default function AgentResourceAccessShowPage({ type }) {
...
     if (!podRoot || !decodedIri) return;
-    // FIXME: write tests for this
-    /* istanbul ignore next */
-    fetch("https://access.pod.inrupt.com/graphql/", {
-      method: "POST",
-      headers: { "Content-type": "application/json" },
-      body: JSON.stringify({ query }),
-    })
+    session
+      .fetch("https://access.pod.inrupt.com/graphql/", {
```
ThisIsMissEm, line 124: As far as I know, this endpoint is going away according to Aaron.

ThisIsMissEm (commented):
This looks okay to me, but we should definitely be careful about writing hooks that contain error state but do not pass that back to the caller, as this will likely result in bugs.

chelseapinka (approved):

PR #504, opened 2022-08-02
removed webID from accept deny funcs

solid-akb:
## Description
Clean up of code. 
## Changes
The SDK removed the argument of the resourceOwner from their code. This change is now reflected in PodBrowser where we remove the resourceOwner from the argument list when calling `approveAccessRequest` and `denyAccessRequest` 
## Testing
Start the request access flow. 
Get re-routed to PodBrowser (make sure the reroute brings you to localhost PB and not prod PB, you can update the URL to reflect this too)
Accept Request. 
Login to PodBrowser as the requesting user. 
In the upper right corner where it says "change pod", put in the resource URI that you requested access to. 
CHECK IF YOU HAVE ACCESS. If yes it worked.  
Start the request access flow again, use the same resource that you just confirmed you have access to. 
Get re-routed to PB. Again confirm you are in localhost and not podbrowser.inrupt.com
CLick Deny
Login to PodBrowser as the requesting user. 
In the upper right corner where it says "change pod", put in the resource URI that you requested access to. 
CHECK THAT YOU DON"T HAVE ACCESS. If you don't, it worked.  
## Commit checklist

### DIFF
--- a/components/accessRequestForm/index.jsx
+++ b/components/accessRequestForm/index.jsx
@@ -133,7 +133,6 @@ export default function AccessRequestForm({
       const expirationDate = selectedDate ? new Date(selectedDate) : null;
       expirationDate?.setUTCHours(23, 59, 59, 0);
       const signedVc = await approveAccessRequest(
-        session.info.webId,
         accessRequest,
         {
           requestor,
@@ -180,10 +179,7 @@ export default function AccessRequestForm({
 
   useEffect(() => {
     const handleDenyAccessRequest = async () => {
-      const signedVc = await denyAccessRequest(
-        session.info.webId,
-        accessRequest
-      );
+      const signedVc = await denyAccessRequest(accessRequest);
       if (signedVc) {
         await router.push(
           `${redirectUrl}?${GRANT_VC_URL_PARAM_NAME}=${getVcId(signedVc)}`


VirginiaBalseiro:
I have one question because I am out of the loop after being away :) [we were relying on the resource owner WebID to decide whether to display a request to a user or not](https://github.com/inrupt/pod-browser/blob/f6b4bc435c5cabde7841b6de698f9f55ff5699ce/components/accessRequestForm/index.jsx#L230) (because requests should only be accessible to the resource owner). How are we planning to handle that from now on?

solid-akb:
Removing the webId from the accept and deny function isn't actually changing the Access Request. The webId of the resource owner is still in the Access Request object, but we are just removing it from being an additional parameter when the user accepts/denies the request. 

We are getting the `resourceOwnerWebId` variable from the [`getDataSubjectWebId` function](https://github.com/inrupt/pod-browser/blob/5644a5779868cb38087686d2d0b8556c364c2ccc/src/models/access/request/index.js#L68) which is looking at the Access Request itself so it shouldn't be dependent on approveAccessRequest or denyAccessRequest being called.

chelseapinka (approved):

VirginiaBalseiro:
Tested, everything looks to be working correctly 👍

VirginiaBalseiro (approved):

PR #507, opened 2022-08-29
download resource

VirginiaBalseiro:
## Description

When navigating to a resource to which a user has access (by any means, including authenticated agent) but does not have access to parent container, PodBrowser tried to load the parent container and incorrectly rendered an error saying the user has no access to the resource.

This PR changes that by checking what the user's access is to both resource and parent container, and either triggering a download and isplaying a download message, or rendering an error if user truly does not have access to the resource.

## Testing

1. Log in as User A, share a resource with Anyone
2. Log in as User 2, try to navigate to the resource via the Pod Navigator
3. Verify that a download starts and a download message is displayed
4. Verify that the no access error is displayed by repeating the above steps with a private resource 

## Commit checklist

### DIFF
--- a/components/container/index.jsx
+++ b/components/container/index.jsx
@@ -60,6 +60,8 @@ import { locationIsConnectedToProfile } from "../../src/solidClientHelpers/profi
 import { isContainerIri } from "../../src/solidClientHelpers/utils";
 import DownloadLink, { downloadResource } from "../downloadLink";
 import styles from "../resourceDetails/styles";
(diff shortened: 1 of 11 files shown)


KyraAssaad (changes requested):
Just some CSS changes plz :) 

1. Turn "Downloading resource" into an H1 and add "..." at the end
2. Add `text-align: center` and `padding-top: 160px' to the span
3. Add a `br` between the body text and the button/link

image

KyraAssaad (approved):
Looks good!

chelseapinka (approved):

PR #514, opened 2022-11-01
pass fetch fn to deny function

VirginiaBalseiro:
## Description

This PR fixes the issue where `denyAccessRequest` was not being called with an options object, causing it to fail. PodBrowser is still at access grants `1.0.1`, so this fixes the issue at the current version and we can look into upgrading separately, as it introduces other challenges.


ThisIsMissEm (approved):
Is there a unit test we could use to verify that fetch was passed to that function? e.g., by mocking it? (if tests are setup for this code)

ThisIsMissEm (commented):
Nice one on the tests!

PR #524, opened 2023-01-05
Update to node 16

solid-akb:
## Description
This PR updates `.nvmrc` and github workflow so that folks use node 16 when running PB locally.


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: 14
+          node-version: 16
```
ThisIsMissEm, line 16: We can actually do: 
```suggestion
          node-version-file: '.nvmrc'
```

Though, now that everyone uses `fnm`, we can rename `.nvmrc` to `.node-version` which is a bit clearer, so the change would then be:
```suggestion
          node-version-file: '.node-version'
```

ThisIsMissEm (changes requested):

ThisIsMissEm:
You'll also want to ensure that the workflow action dependencies are up to date (so, e.g., `actions/setup-node@v3` instead of `@v2`), I know we did this, I'm not sure if it's in a different branch.

ThisIsMissEm:
@solid-akb upgrading `@datapunt/matomo-tracker-react` to `^0.5.1` should resolve the peer dependency conflict and allow for upgrading react to 18.

ThisIsMissEm:
@solid-akb I think we should either close this or mark it as a draft, what do you think?

PR #527, opened 2023-01-05
chore: correctly negotiation content-type for Client Identifier document

ThisIsMissEm:
## Description

This brings pod-browser inline with the specification. We will advertise as `application/ld+json` if either: the Accepts header allows `application/ld+json` or if the Accepts header indicates some other random content type (e.g., text/html, then we serve as json, because there is no HTML representation, but it is handy to be able to view a client identifier document in your browser)

## Changes

- Removes nock from the API route tests
- Switches to using correct request injection testing of the handler
- Changes the endpoint to return a 400 bad request if the request method is NOT a GET request (there's no reason a service should do anything else)
- Supports sending back `application/ld+json` content-type if requested via Accept header — the default *has changed* from `application/json` to `application/ld+json` as this is still compatible with deployments of ESS 2.0.

## Testing

```npm run test -- pages/api/app.test.js```

Alternatively, try logging in in the preview environment with both ESS PodSpaces and use.id (`idp.use.id`)

## Commit checklist

## Interested parties

@chelseapinka @tomhgmns (external, use.id) @solid-akb

### CODE REVIEW

**Comment by ThisIsMissEm** [comment]
Testing results:
- ESS PodSpaces ✅ 
- ESS Dev-Next ✅ 
- NSS (inrupt.net) ✅  — it doesn't verify the client identifier document
- CSS (solidcommunity.net) ✅  — iirc, this is using CSS, but it doesn't verify the client identifier document
- use.id (idp.use.id) ❌  — appears to not send an Accept header, meaning it receives `application/json` when it wants `application/ld+json`

File: pages/api/app.js [hunk 1]
```diff
@@ -37,7 +38,33 @@ function buildAppProfile(hostname, clientId) {
 }
 
 export default function handler(req, res) {
+  if (req.method !== "GET") {
+    return res.status(400).send("Bad Request");
```
NSeydoux, line 42:
```suggestion
    return res.status(405).send("Method not allowed");
```

pages/api/app.test.js
```diff
@@ -19,52 +19,148 @@
...
-  let res;
-
-  beforeEach(() => {
-    req = createRequest();
-    res = createResponse();
-    // eslint-disable-next-line func-names
-    res.status = jest.fn(function () {
-      return this;
+  describe("errors", () => {
+    it("responds with 400 Bad Request if the request method is not GET", async () => {
```
NSeydoux, line 53: See my suggestion to respond with 405

pages/api/app.js
```diff
@@ -37,7 +38,33 @@ function buildAppProfile(hostname, clientId) {
...
+  }
+
   const clientId = `https://${req.headers.host}/api/app`;
   const hostname = `https://${req.headers.host}/`;
-  res.status(200).json(buildAppProfile(hostname, clientId));
+
+  const acceptedType = accepts(req).types([
+    "application/json",
```
NSeydoux, line 49: It isn't obvious that the first type here is returned in the absence of an `Accept` header (which I find a bit odd, as opposed to returning `false` like it does when the negotiation fails). I think it would be worth documenting in a comment.

ThisIsMissEm:
@NSeydoux ended up also moving some logic from the constants directory to a separate file, as the api/app unit tests were complaining a bit about code coverage of the logic imported by the constants which relied on data in the window/DOM.

ThisIsMissEm:
Testing results:

- ESS PodSpaces ✅
- ESS Dev-Next ❌  — currently having issues due to a bad upgrade
- NSS (inrupt.net) ✅ — it doesn't verify the client identifier document
- CSS (solidweb.me) ✅
- use.id (idp.use.id) ✅  — fixed, added response_types of code, even though this is a default, use.id [doesn't respect that](https://gitlab.com/vincenttunru/penny/-/issues/25#note_1242624516), which is their bug.

NSeydoux (approved):

ThisIsMissEm:
Testing results:

ESS PodSpaces ✅
ESS Dev-Next ✅ 
NSS (inrupt.net) ✅ — it doesn't verify the client identifier document
CSS (solidweb.me) ✅
use.id (idp.use.id) ✅ — fixed, added response_types of code, even though this is a default, use.id [doesn't respect that](https://gitlab.com/vincenttunru/penny/-/issues/25#note_1242624516), which is their bug.

PR #532, opened 2023-01-31
feat: add use.id to default providers

ThisIsMissEm:
## Description

This adds use.id to the list of providers that show up in the dropdown menu when selecting "other provider" when logging in, this makes sense, as use.id don't actually advertise their `idp.use.id` URI anywhere in their dashboard, so it's hard to discover the correct URI to use when logging in with Pod Browser.

## Changes

Alongside this change, I've also fixed the development client identifier document that we use, which I'd broken by accident in #527 

Before:
Screenshot 2023-01-31 at 19 48 33

After:
Screenshot 2023-01-31 at 19 49 04

## Testing

Have manually tested on localhost.

## Commit checklist

- [ ] Includes tests to ensure functionality and prevent regressions. (n/a, we don't have coverage for this)


.env.example
```diff
@@ -1 +1 @@
-NEXT_PUBLIC_DEV_CLIENT_ID=
+NEXT_PUBLIC_DEV_CLIENT_ID=https://storage.inrupt.com/87048fc7-b553-4c86-95de-633e1675f0bd/clientid.jsonld
```
chelseapinka, line 1: Do we need this value populated in the example env?

ThisIsMissEm: Normally you'd only have `.env.example` checked in, so it makes sense to have the value there. Though for some reason we check in `.env.development` too.

chelseapinka (approved):

